PR #7, opened 2015-03-20
Update installation.rst

gsirin:
This was a little bit confusing during manual installation...


joschi:
Thanks! We will fix that for the documentation of Graylog 1.0.x but the cluster name will be changed to `graylog` for all subsequent releases.

PR #11, opened 2015-04-09
rsyslog has a built-in template implementing RFC5424

quakehead:


joschi:
Yes, indeed. It's the same format being required by syslog4j (http://syslog4j.org/docs/javadoc/org/productivity/java/syslog4j/impl/message/structured/StructuredSyslogMessage.html) and defined in https://tools.ietf.org/html/draft-ietf-syslog-protocol-23#section-6

Thanks!

PR #22, opened 2015-06-10
Document the templates for the email alert callback

bernd:
Adds documentation for the email alert callback.

Refs Graylog2/graylog2-server#1156
Refs Graylog2/graylog2-web-interface#1355


pages/streams.rst
```diff
@@ -118,6 +118,46 @@ Alarm callbacks are a list of events that are being processed when an alert is t
...
+``check_result``
+  The check result object for this stream.
+
+  * ``check_result.triggeredCondition`` string representation of the triggered alert condition
+  * ``check_result.triggeredAt`` date when this condition was triggered
+  * ``check_result.resultDescription`` text that describes the check result
+``backlog``
+  List of message objects. Can be used to iterate over the messages via ``foreach``.
+
+  Each message in the list has the following fields.
```
joschi: We should add a hint to the `message` object (and that the `toString()` representation is being used when not qualifying it further) so that this is consistent with the other variables (e. g. `stream` or `check_result`).

bernd:
Updated after comment from @joschi.

joschi:
LGTM. :+1:

PR #25, opened 2015-08-06
Documentation only.  Add notes on search syntax peculiarities.

jputney:
Documentation change only.  Add notes on case sensitive AND/OR/NOT and on wildcards only working on message, full message and source by default.


joschi:
Thanks!

PR #30, opened 2015-10-23
sphinx and theme instruction for Fedora

tjyang:
Hi
I am on Fedora platform.  "make html;make latex" work without issue. LaTeX dependency instruction for "make latexpdf" will come later.


lennartkoopmann:
Thank you!

PR #34, opened 2015-11-02
New FAQs

ilynnej:


pages/faq.rst
```diff
@@ -5,61 +5,125 @@ Frequently asked questions
...
 
-The MongoDB dependency of Graylog is there to store metadata that is not log data.
-None of your messages is ever stored in Graylog but for example user information
-or stream rules are. This is why you should not expect much load on MongoDB and
-thus don't have to worry too much about scaling it. It will just run aside your
-`graylog-server` processes and take almost no resources in our recommended setup
-architectures.
+Graylog uses MongoDB to store your configuration data, not your log data. Only metadata is stored, such as user information or stream configurations. None of your log messages are ever stored in MongoDB. This is why MongoDB does not have a big system impact, and you won’t have to worry too much about scaling it. With our recommended setup architecture, MongoDB will simply run alongside your graylog-server processes and use almost no resources.
+
+We have plans to introduce a database abstraction layer in the future. This will give you the flexibility to run MongoDB, MySQL or any other database to store metadata.
```
joschi: I think this shouldn't be mentioned here as long as we don't have a (rudimentary) roadmap for the replacement.

lennartkoopmann: We should mention it because it is a concern of many users and we can show that we are aware of this being not optimal. It is good to have it in there as long as we don't mention *when*.

lennartkoopmann:
Also merged into 1.2.

PR #37, opened 2015-11-23
Cleaned up 1.3 getting started docs

zexxon:
Minor edits in general and tweaks to getting messages in section.


SamParikh:
I'm good, thanks.

PR #55, opened 2016-02-22
Emphasize memory requirement

sarciszewski:
I lost several days of painful debugging to overlooking this detail

### DIFF
--- a/pages/installation/aws.rst
+++ b/pages/installation/aws.rst
@@ -11,7 +11,7 @@ Usage
 -----
 
   * Click on *Launch instance* for your AWS region to start Graylog into.
-  * Choose an instance type with at least 4GB memory
+  * Choose an instance type **with at least 4GB memory**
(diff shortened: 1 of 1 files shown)


sarciszewski:
Also:

https://github.com/Graylog2/documentation/blob/master/pages/faq.rst#isnt-java-slow-does-it-need-a-lot-of-memory

> This is a concern that we hear from time to time. We understand Java has a bad reputation from slow and laggy desktop/GUI applications that eat a lot of memory. However, we are usually able to prove this assumption wrong. Well written Java code for server systems is very efficient and does not need a lot of memory resources.

+ 

> at least 4GB memory

One of these things is not like the other.

mariussturm:
Thanks!
You have to understand that on the appliances all services run on a single virtual machine. So MongoDB, Elastichsearch, Graylog itself, Etcd and other services are all calculated into the 4GB.

PR #57, opened 2016-02-26
Changed link text

bodiam:
Made text a bit easier to read without having to hover over the link the see where 'here' takes you.


joschi:
Makes sense. Thanks!

bodiam:
It's only a small contribution, I hope to make better ones soon. Thanks for merging it so quickly!

bernd:
> It's only a small contribution, I hope to make better ones soon. Thanks for merging it so quickly!

:heart:

PR #63, opened 2016-03-17
Update graylog_ctl.rst to show correct upgrade link

renapu:
changed the upgrade link to the correct one, otherwise you will get only html code with the wget command.


mariussturm:
Thanks, messed that up yesterday!

PR #66, opened 2016-03-29
How to avoid ES split-brain and how to avoid shard shuffling

mikkolehtisalo:
As per https://github.com/Graylog2/graylog2-server/issues/533.


bernd:
Thank you very much! :smiley:

PR #68, opened 2016-04-12
More information about how to manage LDAP referral issues

mikkolehtisalo:
I have previously used global catalog in production with no issues. Should imho be mentioned as an option.


bernd:
Thank you!

PR #70, opened 2016-04-23
Missing trailing slash in configuration

sjoerdmulder:
Otherwise API gives only 404's


joschi:
@sjoerdmulder Thanks!

PR #71, opened 2016-04-25
Working configuration for Apache 2.4 as reverse https proxy

mvazquezc:
Working with RC1

### DIFF
--- a/pages/configuring_webif.rst
+++ b/pages/configuring_webif.rst
@@ -182,6 +182,31 @@ If you want to use nginx to proxy access to a Graylog server, you have several o
     }
   }
 
+Apache:
+------- 
+
+For the following samples we are assuming that your Graylog instance is running on ``graylog.internal.example.org`` using the default ports of 12900 for the REST API and 9000 for the web interface. SSL is disabled for both. You want to expose the Graylog web interface as ``https://graylog.example.org``. 
+
+**URI Configs in Graylog server conf**::
+
+   rest_listen_uri = http://127.0.0.1:12900/
+   web_listen_uri = http://127.0.0.1:9000/
+   web_endpoint_uri = https://graylog.example.org/api/
+ 
+**REST API and Web Interface on one port (using HTTPS/SSL)**::
+   Listen 443
+   <VirtualHost *:443>
+       ServerName graylog.example.org
+       Your SSL config
+       <Location />
+           ProxyPass http://127.0.0.1:9000/
+           ProxyPassReverse http://127.0.0.1:9000/
+       </Location>
+       <Location /api/>
+           ProxyPass http://127.0.0.1:12900/
+           ProxyPassReverse http://127.0.0.1:12900/
+       </Location>
+   </VirtualHost>
 
 AWS Elastic Load Balancing:
 ---------------------------
--- a/pages/configuring_webif.rst
+++ b/pages/configuring_webif.rst
@@ -191,14 +191,15 @@ For the following samples we are assuming that your Graylog instance is running 
 
    rest_listen_uri = http://127.0.0.1:12900/
    web_listen_uri = http://127.0.0.1:9000/
-   web_endpoint_uri = https://graylog.example.org/api/
- 
+   
+   
 **REST API and Web Interface on one port (using HTTPS/SSL)**::
    
    Listen 443
    <VirtualHost *:443>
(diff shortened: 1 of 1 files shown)


pages/configuring_webif.rst
```diff
@@ -182,6 +182,31 @@ If you want to use nginx to proxy access to a Graylog server, you have several o
     }
   }
 
+Apache:
+------- 
+
+For the following samples we are assuming that your Graylog instance is running on ``graylog.internal.example.org`` using the default ports of 12900 for the REST API and 9000 for the web interface. SSL is disabled for both. You want to expose the Graylog web interface as ``https://graylog.example.org``.
```
dennisoelkers: Could you please emphasize that the Graylog server is running on the same host as Apache? Makes it clearer for people adapting the config.

pages/configuring_webif.rst
```diff
@@ -182,6 +182,31 @@ If you want to use nginx to proxy access to a Graylog server, you have several o
...
+
+   rest_listen_uri = http://127.0.0.1:12900/
+   web_listen_uri = http://127.0.0.1:9000/
+   web_endpoint_uri = https://graylog.example.org/api/
+ 
+**REST API and Web Interface on one port (using HTTPS/SSL)**::
+   Listen 443
+   <VirtualHost *:443>
+       ServerName graylog.example.org
+       Your SSL config
```
dennisoelkers: Please add a comment sign to make clear that this needs to be modified after copy & pasting :)

pages/configuring_webif.rst
```diff
@@ -182,6 +182,31 @@ If you want to use nginx to proxy access to a Graylog server, you have several o
...
+Apache:
+------- 
+
+For the following samples we are assuming that your Graylog instance is running on ``graylog.internal.example.org`` using the default ports of 12900 for the REST API and 9000 for the web interface. SSL is disabled for both. You want to expose the Graylog web interface as ``https://graylog.example.org``. 
+
+**URI Configs in Graylog server conf**::
+
+   rest_listen_uri = http://127.0.0.1:12900/
+   web_listen_uri = http://127.0.0.1:9000/
+   web_endpoint_uri = https://graylog.example.org/api/
```
dennisoelkers: I would prefer doing this with the `X-Graylog-Server-URL`-header, because it makes the setup more flexible (you can point multiple proxies to it, exposing the web interface under different URLs). What do you think?

dennisoelkers:
Thanks much for the write up! Much appreciated!

pages/configuring_webif.rst
mvazquezc: Done!

pages/configuring_webif.rst
mvazquezc: Done!

pages/configuring_webif.rst
mvazquezc: Hi Dennis,

I agree with you, this way configuration will be more flexible.

Working on it. Once change is done and tested I will commit changes again.

Thank you,

dennisoelkers: Thanks again!

mvazquezc: Commit done :) 

You're welcome. Thank you for such a great product :D

ross-bragg:
Hey @mvazquezc- I don't know if this is a quirk of my setup or if it's an actual error, but I had to move the `RequestHeader set X-Graylog-Server-URL` inside the `<Location />` on my apache config. If you apply the header to both sets of requests, you get a lot of 405s and you can't log in (if you can even get to the login screen).

ross-bragg:
Hey @mvazquezc - I don't know if this is a quirk of my setup or if it's an actual error, but I had to move the `RequestHeader set X-Graylog-Server-URL` inside the`<Location />` on my apache config. If you apply the header to both sets of requests, you get a lot of 405s and you can't log in (if you can even get to the login screen).

joschi:
@ross-bragg While I'm not sure that it's an actual error, I think moving the `RequestHeader` directive into the `<Location>` container for proxying the web interface is clearer than before. I've changed that in 7cf09097b2a08fe83e1459d1c34c45a54d5cfd15.

mvazquezc:
In my setup it was working fine (Apache 2.4.6), but I agree with @joschi, moving RequestHeader inside <Location /> is a good idea.

PR #74, opened 2016-04-28
Fixed typo in graylog_ctl.rst

hjbyt:


joschi:
@hjbyt Good catch, thank you!

PR #76, opened 2016-05-01
Tls ca check

mikkolehtisalo:
Probably good to document how to approach TLS handshake issues. Should be enough for issues like Graylog2/graylog2-server#1668 - unless if a whole HOWTO should be provided...


joschi:
@mikkolehtisalo Thank you very much!

PR #84, opened 2016-05-31
Add Graylog Enterprise documentation

bernd:
- Add captions to the navigation sidebar to separate OSS and Enterprise
  docs
- Point Archive installation notes to the Enterprise setup page
- Add Enterprise setup page


index.rst
```diff
@@ -13,6 +13,7 @@ Contents:
 
 .. toctree::
    :maxdepth: 2
+   :caption: Graylog Server
```
edmundoa: Maybe call this caption simply Graylog? In the end we describe more than just the server (web interface, and some plugins)

pages/enterprise/setup.rst
```diff
@@ -0,0 +1,115 @@
...
+      - Directory
+    * - :doc:`/pages/installation/virtual_machine_appliances`
+      - ``/opt/graylog/plugins/``
+    * - :doc:`/pages/installation/operating_system_packages`
+      - ``/usr/share/graylog-server/plugin/``
+    * - :doc:`/pages/installation/manual_setup`
+      - ``/<extracted-graylog-tarball-path>/plugin/``
+
+Make sure to install the enterprise plugin JAR files alongside the other Graylog plugins.
+Your plugin directory should looke similar to this after installing the enterprise plugins.
```
edmundoa: Typo in `looke`

pages/enterprise/setup.rst
```diff
@@ -0,0 +1,115 @@
...
+
+::
+
+  $ tar -tzf graylog-enterprise-plugins-1.0.0.tgz
+  graylog-enterprise-plugins-1.0.0/plugin/graylog-plugin-archive-1.0.0.jar
+  graylog-enterprise-plugins-1.0.0/plugin/graylog-plugin-license-1.0.0.jar
+
+Depending on the Graylog setup method you have used, you have to install the plugins into different locations.
+
+.. list-table:: Plugin Installation Locations
```
joschi: This information would also be useful in the normal installation docs.

edmundoa: We should also indicate that the plugins need to be installed in all Graylog servers inside the cluster. I think that is mentioned somewhere in the docs, but we should remark it here as well.

bernd:
Addresses all comments.

bernd:
Added missing includes directory.

edmundoa:
LGTM 👍

Issue #87, opened 2016-06-08
Add a section about Securing Graylog


mikkolehtisalo:
It might be good to refactor the documentation a bit, and to divide information a bit like
- Installing: "What takes to get it running"
- Configuring: "How to configure Graylog to accomplish stuff"
- Securing: "How to harden the setup"

Anyways, _Securing Graylog_ should collect together for example
- Enabling audit logging of usage (I will make a pull request of this very shortly)
- Network traffic profile (internal in cluster, outside, firewall requirements, etc)
- Configuring for cryptographic compliance (this part exists, it's brand new)
- Configuring "real" certificates for REST API and inputs
- How to add a legal disclaimer to login page, and to footer (I am using Apache's mod_substitute, but it's a hack)
- How to configure authentication (the pluggable authentication is in development, I'm especially interested in kerberos/GSSAPI and TLS client certificate authentication)
- What security maintenance tasks should be performed in production environments
- Other pieces of security related configuration that are all around the current documentation set

joschi:
@mikkolehtisalo Thanks, that's a very good idea!

jalogisch:
@mikkolehtisalo first part should be covered with https://github.com/Graylog2/documentation/commit/7e997f496fd33085ee16d035d85d35fe7bbc32ac what is your opinion on this?

mikkolehtisalo:
@jalogisch  Some good things there! In any case, many companies and such raise the questions "is X secure", and "how to harden X to achieve compliance Y" as a separate step in the process of approving  systems for use. Being able to find and verify all necessary (sometimes even small) details is important. 

Also, my thinking is that the documentation is growing. It needs more structure, or it becomes messy. (Mostly separating basic installation from extra configuration, which can vary greatly, but this also goes for some security topics that are strictly not mandatory...)

jalogisch:
@mikkolehtisalo i see your points. i personal had started to restructure the documentation a little with this https://github.com/Graylog2/documentation/pull/96 

Divide between simple and complex setups is not that easy as we would need to rewrite most parts of the documentation and change the logical structure. But we could add some points to the documentation to have them available. 

My idea would be to add:
- how to setup and use mongo in HA (in short what needs to be done within mongo and what needs to be configured on graylog)
- how a possible HA setup can look and how this can be secured best by design 

[...]

PR #88, opened 2016-06-08
Instructions on how to configure RestAccessLog

mikkolehtisalo:
Related to #87.

I could not find this from the documentation. Please feel free to refactor & pull. :)

### DIFF
--- a/pages/securing.rst
+++ b/pages/securing.rst
@@ -1,0 +1,58 @@
+.. _securing:
+
+****************
+Securing Graylog
+****************
+
+Logging user activity
+=====================
+
+Graylog follows microservices architecture model where the user interface retrieves all data via a collection of REST APIs. Thus logging relevant user activity is a simple matter of enabling built-in feature called RestAccessLogFilter. It catches all requests to REST APIs, and produces an access log augmented by user information.
+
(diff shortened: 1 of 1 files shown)


pages/securing.rst
```diff
@@ -0,0 +1,58 @@
+.. _securing:
+
+****************
+Securing Graylog
+****************
+
+Logging user activity
+=====================
+
+Graylog follows microservices architecture model where the user interface retrieves all data via a collection of REST APIs. Thus logging relevant user activity is a simple matter of enabling built-in feature called RestAccessLogFilter. It catches all requests to REST APIs, and produces an access log augmented by user information.
```
joschi: It's just a client-server model and doesn't have anything to do with micro services (Bingo! 🎉).

mikkolehtisalo: You are correct. The way services are designed, it really resembles more client-server model. It's just twisted to talk about client-server when the actual user facing application is entirely dependent on the server - and not just the data. But I guess the browser can be the client, hehe. Anyways, I changed that part. :)

joschi:
@mikkolehtisalo Thanks a lot!

PR #90, opened 2016-06-09
Update to reflect 2.0 changes

varnav:
graylog.conf in now server.conf
Multicast discovery has been removed from Elasticsearch 2.x
Dynamic scripting is not an issue in 2.0

### DIFF
--- a/pages/configuring_es.rst
+++ b/pages/configuring_es.rst
@@ -47,51 +47,13 @@ Cluster Name
 
 You need to tell ``graylog-server`` which Elasticsearch cluster to join. The Elasticsearch cluster default name is *elasticsearch*
 and configured for every Elasticsearch node in its ``elasticsearch.yml`` configuration file with the setting ``cluster.name``.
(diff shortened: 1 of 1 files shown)


joschi:
@varnav Thank you for your contribution!

> graylog.conf in now server.conf

Unfortunately that depends on the installation method. In fact, the configuration file can have an arbitrary name, so we might want to rephrase that entirely.

> Multicast discovery has been removed from Elasticsearch 2.x

While this is partly true (Multicast discovery still exists as an plugin), setting `elasticsearch_discovery_zen_ping_unicast_hosts` is mandatory in Graylog 2.x (since the defaults settings of Elasticsearch 2.x also changed). We'll have to rewrite that paragraph accordingly.

> Dynamic scripting is not an issue in 2.0

While dynamic scripting is disabled by default in Elasticsearch 2.x, I'd like to keep this section in our documentation.

joschi:
Feel free to address my comments, otherwise you can close this PR (or I will do so later).

PR #93, opened 2016-06-10
Fixed a highlighting issue

walked:
sudo parted -a optimal /dev/sdb mklabel gpt was not properly highlighted as syntax


joschi:
@walked Thanks!

PR #97, opened 2016-06-14
Fixing broken set-external-ip command in aws.rst

tzachyrm:


123dev:
Thanks
sorry for missing the fix.

Issue #10, opened 2015-04-07
Document backup procedures


henrikjohansen:
I would be beneficial to have an official description of how to backup the graylog configuration that is persisted into mongodb ...

jalogisch:
@henrikjohansen did you worked out your backup already? can you describe?

henrikjohansen:
We dump the mongo db as suggested. I would still like to see something like that build into Graylog as a system job (ie. daily dumps of the configuration in mongo to a file with configurable interval and retention). 

I do not entirely agree with your comments in commit ce6ed82 - when it comes to a **configuration** backup there really only is one way to do so (every system is not special in that regard).

jalogisch:
hej @henrikjohansen i can see your point. 

In your point of view, adding something like https://github.com/micahwedemeyer/automongobackup and give graylog this function as plugin or in core would be recommend?

henrikjohansen:
Absolutely, yes.

jalogisch:
we have the server issue to create backup plugin or similar we do not need to have this issue open in the documentation. 

This task is very specific and the notes already in the docs are sufficient.

PR #114, opened 2016-06-23
added first notes about backup

jalogisch:
this is just the first step to have something said about backup in our documentation.

it does not cover details and leave the people on their own but everyone with good knowlegde can develop backup and restore that fits into his situation.

reference: https://github.com/Graylog2/documentation/issues/10


edmundoa:
LGTM 👍

Issue #115, opened 2016-06-23
Data storage location clartification


sandynomad:
On the page http://docs.graylog.org/en/2.0/pages/configuration/graylog_ctl.html there is a section "All data is stored in one directory /var/opt/graylog/data.".

On page http://docs.graylog.org/en/2.0/pages/faq.html there is a section "Does Graylog encrypt log data?
All log data is stored in Elasticsearch.".

The default elasticsearch storeage location on a C7 installation is in /var/lib/elasticsearch/graylog.

What is expected to be stored at /var/opt/graylog/data? (On our fresh install after running for 48 hours nothing has been stored in this directory).

How much storage space will elasticsearch use? (After 48 hours /var/lib/elasticsearch/graylog has > 300MB)

jalogisch:
Dear @sandynomad,

the information you found in the first page is used inside the OVA Image. I will try to clarify this in the Documentation.

In your case all data will be stored in your elasticsearch folder and the size it can gain depends on the input and the time you want to keep your data.

PR #117, opened 2016-06-24
little correction on storage location

jalogisch:
to make it more clear where /var/opt/graylog/data is use
ref: https://github.com/Graylog2/documentation/issues/115


mariussturm:
LGTM :+1:

PR #122, opened 2016-07-05
HTTPS troubleshooting tip

PaulDMendoza:


joschi:
As mentioned in Graylog2/graylog2-server#2456, the domain you're using in `rest_listen_uri`, `rest_transport_uri`, and `web_listen_uri` must be resolvable on the system running Graylog. This doesn't have anything to do with HTTPS at all.

Additionally, it would be better if the configured domain name(s) could be resolved by the system's DNS resolver rather than only be added to the local `/etc/hosts` file.

PR #123, opened 2016-07-05
PFX and hints on SSL

PaulDMendoza:
It took me a while to realize a PFX file was actually a PKCS#12 file so I figured documentation on that would help. Also, the hosts file work around should be documented.


PaulDMendoza:
To be clear, I've documented the hosts file work around here.

joschi:
@PaulDMendoza Thank you for your contribution! I've cherry-picked the part about converting PFX files from this PR and discarded the rest.

PR #124, opened 2016-07-07
little updates on installation

jalogisch:
Added some additional information on the configuration/setup

### DIFF
--- a/pages/installation/operating_system_packages.rst
+++ b/pages/installation/operating_system_packages.rst
@@ -131,5 +131,6 @@ Step-by-step guides
 Feedback
 --------
 
-Please open an `issue <https://github.com/Graylog2/fpm-recipes/issues>`_ in the `Github repository <https://github.com/Graylog2/fpm-recipes>`_ if you
-run into any packaging related issues. **Thank you!**
+Please open an `issue in the Github repository for the operating system packages <https://github.com/Graylog2/fpm-recipes>`__ if you
+run into any packaging related issues. If this Documentation is confusing or left questions to you please open an `issue in the Github repository for the documentation <https://github.com/Graylog2/documentation/issues>`__.
+
--- a/pages/installation/operating_system_packages.rst
+++ b/pages/installation/operating_system_packages.rst
@@ -128,10 +128,11 @@ Step-by-step guides
    os/ubuntu
    os/centos
 
+
 Feedback
 --------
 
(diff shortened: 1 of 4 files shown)


pages/installation/operating_system_packages.rst
```diff
@@ -131,5 +131,6 @@ Step-by-step guides
 Feedback
 --------
 
-Please open an `issue <https://github.com/Graylog2/fpm-recipes/issues>`_ in the `Github repository <https://github.com/Graylog2/fpm-recipes>`_ if you
-run into any packaging related issues. **Thank you!**
+Please open an `issue in the Github repository for the operating system packages <https://github.com/Graylog2/fpm-recipes>`__ if you
+run into any packaging related issues. If this Documentation is confusing or left questions to you please open an `issue in the Github repository for the documentation <https://github.com/Graylog2/documentation/issues>`__.
```
joschi: Lower case "documentation" (also in all following files of this change set).

Maybe reword to "If you found this documentation confusing or have more questions, please open an issue…"

pages/installation/os/centos.rst
```diff
@@ -85,6 +85,9 @@ The last step is to enable Graylog during the operating system's startup::
   $ sudo systemctl start graylog-server.service
 
 
+.. note:: We higly recommend that you proxy the Graylog Webinterface and API with NGINX/Apache like described in: :ref:`configuring_webif_nginx`
```
joschi: Missing "." after the `ref`erence.

Also, why do we recommend this?

jalogisch: IMHO this Step-by-step is used by people that are not very skilled on linux and for them it is easier to run HTTPS from NGINX/Apache.

pages/installation/os/centos.rst
```diff
@@ -85,6 +85,9 @@ The last step is to enable Graylog during the operating system's startup::
   $ sudo systemctl start graylog-server.service
 
 
+.. note:: We higly recommend that you proxy the Graylog Webinterface and API with NGINX/Apache like described in: :ref:`configuring_webif_nginx`.
```
joschi: I think we should remove this note and rather improve our own HTTPS documentation.

Otherwise we could as well remove HTTPS support from Graylog and always recommend using a 3rd party reverse proxy.

kroepke: same as my comment in https://github.com/Graylog2/documentation/pull/125#issuecomment-231743893

PR #125, opened 2016-07-07
added note about single node setup

jalogisch:
recommend nginx/apache as frontend on HTTPS with single node setup


joschi:
I think we shouldn't add this note and rather improve our own HTTPS documentation if necessary.

Otherwise we could as well remove HTTPS support from Graylog and always recommend using a 3rd party reverse proxy.

kroepke:
I agree with @joschi, I see the benefit of HTTPS termination mostly for multi-node setups (ignoring other reasons for proxies for the moment, such as implementing security policies etc).

If we want to add a note here, I'd simply state that a proxy config is also a possibility but not steer people away from the built-in support.
Maybe something like a "I want HTTPS" section where we list the possible paths?

PR #127, opened 2016-07-12
Change Present Progressive tense to Present Simple

colinangusmackay:
The present simple tense makes more sense in the context of this paragraph. The present progressive (aka present continuous) doesn't fit the style of the rest of the text.


jalogisch:
thank you @colinangusmackay

PR #128, opened 2016-07-12
added multi node setup

jalogisch:
this is just the first writing on multi node setup of graylog in bigger environments.


jalogisch:
thanks for the wording clean up @joschi - in my opinion this can be merged!

PR #141, opened 2016-07-29
added metric/monitoring to faq

jalogisch:
after we had clarified this once, this should be documented.

### DIFF
--- a/pages/faq.rst
+++ b/pages/faq.rst
@@ -164,10 +164,10 @@ I don’t want to use Elasticsearch as my backend storage system – can I use a
 
 You can, but we don’t suggest you do. You will not be able to use our query functionality or our analytic engine on the dataset outside the system. We only recommend another database if you want it for secondary storage.
 
-Can i add a restricted user to Graylog, to inlcude internal Graylog metrics in my Monitoring?
+How can I create a restricted user to check internal Graylog metrics in my monitoring system?
 ---------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


pages/faq.rst
```diff
@@ -164,6 +164,23 @@ I don’t want to use Elasticsearch as my backend storage system – can I use a
 
 You can, but we don’t suggest you do. You will not be able to use our query functionality or our analytic engine on the dataset outside the system. We only recommend another database if you want it for secondary storage.
 
+Can i add a restricted user to Graylog, to inlcude internal Graylog metrics in my Monitoring?
```
joschi: "How can I create a restricted user to check internal Graylog metrics in my monitoring system?"

pages/faq.rst
```diff
@@ -164,6 +164,23 @@ I don’t want to use Elasticsearch as my backend storage system – can I use a
...
+---------------------------------------------------------------------------------------------
+
+You can create a user that has access only to ``/system/metrics`` at the Graylog API, this way it will be possible for you to integrate Graylog internal metrics to your monitoring.
```
joschi: It's also possible without an additional user. Maybe we should include the reasoning behind it (using only restricted users to minimize impact if the credentials leaked).

pages/faq.rst
```diff
@@ -164,6 +164,23 @@ I don’t want to use Elasticsearch as my backend storage system – can I use a
...
+
+POST via API Browser or curl the following to ``/roles`` endpoint of your Graylog API::
```
joschi: Maybe include a fully working `curl` example.

PR #144, opened 2016-08-01
change issue template

jalogisch:
used https://www.talater.com/open-source-templates/#/ to create a new issue template

after @joschi pointed me us to them this morning


joschi:
I think this is still not a good template for the documentation repository.

What are users supposed to write down as expected behavior if the documentation is missing something or something is plain wrong?

joschi:
:+1:

PR #149, opened 2016-08-04
Add documentation for HAProxy configuration (HTTP)

mattcg:
This patch adds an example configuration for a HAProxy reverse proxy in front of Graylog 2.0.


joschi:
Thanks!

PR #150, opened 2016-08-05
show multiple tags in example

0x20h:
That's the format that worked for me. Didn't find an explanation on how to specify multiple tags...


joschi:
@0x20h Thanks!

mariussturm:
@0x20h you can also use the normal yaml list syntax like it's used in the default config: https://github.com/Graylog2/collector-sidecar/blob/master/collector_sidecar.yml#L11

0x20h:
@mariussturm Thanks for the info!

PR #159, opened 2016-08-22
Add basic decorators documentation

edmundoa:
This PR adds some general documentation about decorators, including the two decorators we ship by default on Graylog 2.1.

Until we find a better place for it, I decided to write the decorators documentation alongside other search documentation.


jalogisch:
i like this and i guess - at the moment this is the best place. 

@joschi your opinion?

joschi:
LGTM. 👍

Issue #19, opened 2015-05-22
Document new Kafka inputs in Graylog 1.1.x


joschi:
There are several new Kafka-based inputs in Graylog 1.1.x which need to be documented.

Refs Graylog2/graylog2-server#322 and Graylog2/graylog2-server#1165

jalogisch:
this is already merged - but this case not closed, so i'll close now

joschi:
@jalogisch We're still missing documentation for our Kafka inputs.

PR #161, opened 2016-08-26
added basic amqp and kafka

jalogisch:
if you search in the documentation you did not find valid information about kafka and amqp inputs. this section will present at least some basic information and that they are present.

additional it will fix: https://github.com/Graylog2/documentation/issues/19

### DIFF
--- a/pages/sending_data.rst
+++ b/pages/sending_data.rst
@@ -95,11 +95,14 @@ Use Kafka as transport queue for sending
 
 `Apache Kafka <http://kafka.apache.org>`__ can be used as a transport for different inputs in Graylog. Gelf, Syslog an RAW Inputs can be connected to a `Zookeeper <https://zookeeper.apache.org>`__ that controlls the Kafka Servers. The Topic is filtered by regex and depending on the Input you are able to set various additional settings.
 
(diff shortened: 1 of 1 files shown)


pages/sending_data.rst
```diff
@@ -90,6 +90,19 @@ Try sending an example message using curl::
 Both keep-alive and compression are supported via the common HTTP headers. The server will return a ``202 Accepted`` when the message
 was accepted for processing.
 
+Use Kafka as transport queue for sending
+========================================
+
+`Apache Kafka <http://kafka.apache.org>`__ can be used as a transport for different inputs in Graylog. Gelf, Syslog an RAW Inputs can be connected to a `Zookeeper <https://zookeeper.apache.org>`__ that controlls the Kafka Servers. The Topic is filtered by regex and depending on the Input you are able to set various additional settings.
+
+Examples can be found in the `Graylog Marketplace <https://marketplace.graylog.org/addons?search=kafka>`__.
```
joschi: What examples do you mean here (and in the AMQP paragraph)?

jalogisch: for amqp I would link to syslog via amqp as we do not have such a thing for kafka, i would need to create one ;)

PR #163, opened 2016-08-28
Add the 'share' directory to '.gitignore'

supahgreg:
When initializing a new virtualenv on a Debian-based system 'wheels' binary packages are stored in `<VIRTUAL_ENV>/share/python-wheels`.  This PR simply adds the 'share' directory to '.gitignore'.

**Reference:** Section 3.2 of https://www.debian.org/doc/packaging-manuals/python-policy/ch-module_packages.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,7 @@ _build
 [Ll]ib64
 [Ll]ocal
 [Ss]cripts
+share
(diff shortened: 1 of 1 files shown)


joschi:
@supahgreg Thanks!

Issue #146, opened 2016-08-01
Make alerts section more prominent


lennartkoopmann:
The "Alerts" section is currently buried under the "Streams" page. Given that the alerting is a very important feature, I'd like to move it to its own section in the top level navigation.

edmundoa:
Fixed in #165

PR #164, opened 2016-08-29
made alerts more prominent

jalogisch:
CLOSE:https://github.com/Graylog2/documentation/issues/146

I had rearranged the Alert Section and did a view modifications to reflect the current state. 

Addition i build a subfolder to organise the Stream Documentation, which can later moved complete into this subfolder.


joschi:
This should probably be done on the `2.1` branch instead of the `2.0` branch. Otherwise a bunch of links to the old documentation will break.

jalogisch:
I tried to have cross-links all over - but, I can change and it would be more useful.

jalogisch:
will create a new pr ...

PR #174, opened 2016-09-01
Update URI/network defaults for Graylog 2.1.x

joschi:
The Graylog REST API moved from http://127.0.0.1:12900/ to http://127.0.0.1:9000/api/

In a follow-up PR, we have to re-evaluate the proxy examples (nginx and Apache httpd) because those expect the Graylog REST API and the web interface to run on different ports. Most of them can be simplified for Graylog 2.1.x.

### DIFF
--- a/pages/configuration/web_interface.rst
+++ b/pages/configuration/web_interface.rst
@@ -20,7 +20,7 @@ Single or separate listeners for web interface and REST API?
 Since Graylog 2.1 you have two options when it comes to exposing its web interface:
 
  - Running both on the same port, using different paths (defaulting to ``http://localhost:9000/api/`` for the REST API and ``http://localhost:9000/`` for the web interface), this is the default since 2.1 and is assumed for most parts of the documentation.
- - Running on two different ports (for example ``http://localhost:9000/api/`` for the REST API and ``http://localhost:9000`` for the web interface)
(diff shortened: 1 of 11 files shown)


pages/configuration/web_interface.rst
```diff
@@ -19,8 +19,8 @@ Single or separate listeners for web interface and REST API?
 
 Since Graylog 2.1 you have two options when it comes to exposing its web interface:
 
- - Running both on the same port, using different paths (defaulting to ``http://localhost:12900/`` for the REST API and ``http://localhost:12900/web`` for the web interface), this is the default since 2.1 and is assumed for most parts of the documentation.
- - Running on two different ports (for example ``http://localhost:12900`` for the REST API and ``http://localhost:9000`` for the web interface)
+ - Running both on the same port, using different paths (defaulting to ``http://localhost:9000/api/`` for the REST API and ``http://localhost:9000/`` for the web interface), this is the default since 2.1 and is assumed for most parts of the documentation.
+ - Running on two different ports (for example ``http://localhost:9000/api/`` for the REST API and ``http://localhost:9000`` for the web interface)
```
edmundoa: In this case we should keep the old URI, or at least change one of the ports.

edmundoa:
LGTM 👍

PR #177, opened 2016-09-01
Add Graylog 2.1.0 changelog

joschi:


joschi:
Given that the change log page is getting a bit out of hand, we might want to introduce sub-pages for each Graylog minor release in the future (e. g. `changelog-1.3`, `changelog-2.0`, `changelog-2.1` etc.).

edmundoa:
LGTM 👍

PR #186, opened 2016-09-05
simplify Proxy

jalogisch:
CLOSE: https://github.com/Graylog2/graylog2-server/issues/2772

This changed examples should show the new settings that all is running on the same port.


joschi:
LGTM. 👍

PR #188, opened 2016-09-07
added haproxy multibackends

jalogisch:
i just had tested what type of frontend allows multiple backends with health checks.

nginx only support health-checks in the commercial part.
apache only supports it in versions that are not shipped with the default OS. 

haproxy is the only left and easy to setup


joschi:
LGTM. 👍

PR #189, opened 2016-09-08
Document required changes to default ModSecurity rules

mikkolehtisalo:
ModSecurity is the swiss knife of open source web application firewalls. In case of Graylog, it can make reconnaissance harder by blocking requests of several malicious tools. It could potentially also be used to detect some Graylog specific anomalous activities.

The default rules require some changes, because Graylog's UI and API access break by default. I decided to document the very basics.


jalogisch:
LGTM 👍 

@joschi your opinion?

joschi:
@mikkolehtisalo @jalogisch The contents are fine, but I wonder if this would be better a separate guide akin to https://github.com/Graylog2/graylog-guide-syslog-linux

mikkolehtisalo:
@joschi Yeah there are several other topics as well, and I am still having trouble finding a place for them. The next one on my list being working with SELinux policy for the Graylog main process...

jalogisch:
i'm with you @joschi 

@mikkolehtisalo the best solution would be if you create a guide _secure your graylog setup_ and put all your topics into this. 

So we could link to the guide and you can just push updates without the need to fit them into our documentation. 

will this be a solution for you?

jalogisch (approved):

PR #192, opened 2016-09-13
Remove mentions of database abstraction layer for MongoDB

joschi:
We don't have concrete plans to do that in the short- or mid-term.


joschi:
This PR should be ported to the `2.0` and at least the `1.3` branches, too.

PR #195, opened 2016-09-15
added the reboot

jalogisch:
after testing the update from 2.0.3 OVA and 2.1.0 OVA to 2.1.1 it looks like the reboot should be done.

Without this strange haning processes and errors appear.

This way the system is cleaned.

### DIFF
--- a/pages/configuration/graylog_ctl.rst
+++ b/pages/configuration/graylog_ctl.rst
@@ -238,6 +238,9 @@ Choose the Graylog version you want to install from the `list of Omnibus package
   $ sudo graylog-ctl stop
   $ sudo dpkg -G -i graylog_latest.deb
   $ sudo graylog-ctl reconfigure
+  $ sudo reboot
+
+.. note:: Restart the Server after you had updated should be done to prevent data corruption!
 
 Migrate manually from 1.x to 2.1.x
 ==================================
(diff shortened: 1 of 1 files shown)


pages/configuration/graylog_ctl.rst
```diff
@@ -238,6 +238,9 @@ Choose the Graylog version you want to install from the `list of Omnibus package
   $ sudo graylog-ctl stop
   $ sudo dpkg -G -i graylog_latest.deb
   $ sudo graylog-ctl reconfigure
+  $ sudo reboot
+
+.. note:: Restart the Server after you had updated should be done to prevent data corruption!
```
mariussturm: `Reboot the server after the update procedure to make sure that all services are running in the correct version.` - data corruption sounds like Elasticsearch will break otherwise what is not the case.

123dev:
It would be great if **graylog-ctl reconfigure** keeps the state of the processes as they were instead of blindly starting them all.
it's not hard to do graylog-ctl start when ready.
A lot of times it is not desirable to start the processes without making some manual adjustments after reconfiguration, or as seen above, needing a reboot prior to starting the processes.
If one is not fast enough, to stop the processes as soon as they start, corruption or undesired effects might be unavoidable.

Thanks for listening.

jalogisch:
@123dev could you please create a feature request in https://github.com/Graylog2/omnibus-graylog2/issues

thank you

PR #198, opened 2016-09-19
Adding gosu to plugin download

omercnet:
The current `RUN` command writes the file as `root` so the file isn't readable by `graylog` user
Adding `gosu graylog` to fix permission issues

### DIFF
--- a/pages/installation/docker.rst
+++ b/pages/installation/docker.rst
@@ -130,7 +130,7 @@ In order to add plugins you can build a new image based on the existsing `graylo
 create a new Dockerfile in an empty directory::
 
   FROM graylog2/server:2.1.0-3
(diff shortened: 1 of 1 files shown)


mariussturm:
Sorry, I can not verify this. When I create a new container, e.g. with the Neo4J output plugin it looks like this:

```
root@99b4cdff6597:/usr/share/graylog/plugin# ls -l plugin-output-neo4j-1.2.0.jar
-rw-r--r-- 1 root root 5635729 Sep  8 11:12 plugin-output-neo4j-1.2.0.jar
```

It's owned by root but readable for all users. This plugin is loaded then perfectly fine.

omercnet:
you're right, my bad.
it's chmod'ed 0400 when I was using `ADD` and not `RUN wget`

Issue #200, opened 2016-09-21
Update "Architectural considerations" for Graylog 2.x


joschi:
The description and images in the [Architectural considerations](http://docs.graylog.org/en/2.1/pages/architecture.html) chapter of the documentation is still referring to components of Graylog 1.x and won't work with Graylog 2.x as it is.

Even worse, the images are misleading.

PR #204, opened 2016-09-28
updated architecture

jalogisch:
FIX: https://github.com/Graylog2/documentation/issues/200

I have worked on the Architecure Page, added some links to setup and additional informatio. 

updated the images


joschi:
@jalogisch Please remove the old images (`./images/simple_setup.png` and `./images/extended_setup.png`) if they aren't referenced anywhere else in the documentation.

And I think the new images should have a white background (they have a transparent one right now) to improve readability. Otherwise the grey background of the RTD theme will "shine" through.

joschi:
@jalogisch The images should refer to "Graylog" (or "graylog") and not "graylog-server".

Issue #202, opened 2016-09-21
add https to example


jalogisch:
## Context

People get confused if apache is used for https termination, this need to be covered in the documentation. Including the caveat with self-sign certificates.

http://docs.graylog.org/en/2.1/pages/configuration/web_interface.html#apache-httpd-2-x

PR #212, opened 2016-10-20
add apache ssl example

jalogisch:
CLOSE https://github.com/Graylog2/documentation/issues/202

As I notice that people get confused on this now SSL example for apache

### DIFF
--- a/pages/configuration/web_interface.rst
+++ b/pages/configuration/web_interface.rst
@@ -192,6 +192,27 @@ Apache httpd 2.x
 
     </VirtualHost>
 
+**REST API and Web Interface on one port (using HTTPS/SSL)**::
+
+    <VirtualHost *:443>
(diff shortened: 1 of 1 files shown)


pages/configuration/web_interface.rst
```diff
@@ -192,6 +192,27 @@ Apache httpd 2.x
...
+        ProxyRequests Off
+        SSLEngine on
+        # <- your SSL Settings here!
+
+        <Proxy *>
+            Order deny,allow
+            Allow from all
+        </Proxy>
+
+        <Location />
```
joschi, line 208: Is wrapping the `Proxy*` directives inside the `Location` block still required, given that it points to the root resource (`/`)?

jalogisch: not sure if this is needed. But that was how I did the test on my setup.

PR #218, opened 2016-10-25
Added missing part to start command in First start

ghost:


joschi:
@k1-hedayati Thanks for your contribution to the Graylog documentation!

The original text referred to Upstart's [`start`](http://upstart.ubuntu.com/cookbook/#start) command which obviously doesn't work on Debian Linux.

Since not all Debian-based Linux distributions are using `systemd`, using [`service collector-sidecar start`](http://manpages.ubuntu.com/manpages/xenial/man8/service.8.html) might be preferable over your variant.

Or even better just split up the table row for Debian/Ubuntu Linux to use the correct start command for each distribution.

Mind to change that?

ghost:
Debian from Jessie and Ubuntu from 15.04 use `systemd` by default so currently latest versions of Debian and Ubuntu both use `systemd`. 
We can split tables to prior and after Jessie/15.04, which I personally don't prefer because it means more work for a casual contributer like me :) 
But if you think splitting is necessary please close these PRs and I will merge them into one commit with docs to support both versions.

joschi:
@k1-hedayati 

> Debian from Jessie and Ubuntu from 15.04 use systemd by default so currently latest versions of Debian and Ubuntu both use systemd.

That's nice, but we're supporting Ubuntu 12.04 LTS, 14.04 LTS, and Debian 7 too…

ghost:
Ok so I'll update docs to support them too.

joschi:
@k1-hedayati Anything coming up here or should I close this PR?

ghost:
@joschi I'm gonna create another PR and merge these two, closing this one.

PR #225, opened 2016-11-10
added Information about securing

jalogisch:
Added some notes about secure connection between Sidecar and Graylog and the collectors and Graylog

### DIFF
--- a/pages/collector_sidecar.rst
+++ b/pages/collector_sidecar.rst
@@ -325,6 +325,22 @@ logfiles and ship them with a Filebeat collector to a Beats input that is listen
 
 - Congratulations your collector setup is working now!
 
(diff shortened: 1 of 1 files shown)


jalogisch:
@mariussturm can you please check on that and write your comments.

mariussturm:
Sounds good, thanks!

mchubby:
Might be noteworthy to note you cannot mix secure and unsecure communications on a given Beats Input (IIRC).

mariussturm:
@mchubby I can't think of another protocol that is able to do so but you can create a PR if you think that is helpful.

joschi:
@mariussturm Think `STARTTLS` in SMTP or IMAPv4.

PR #229, opened 2016-11-28
updated Step-by-step guides

jalogisch:
after we got feedback via Mail just a few updates on thie guides

### DIFF
--- a/pages/installation/os/centos.rst
+++ b/pages/installation/os/centos.rst
@@ -92,7 +92,7 @@ The last step is to enable Graylog during the operating system's startup::
   $ sudo systemctl enable graylog-server.service
   $ sudo systemctl start graylog-server.service
 
-The next step is it :ref:`ingest messages <ingest_data>` into your Graylog and extract the messages with :ref:`extractors <extractors>` or use :ref:`the Pipelines <pipelinestoc>` to work with the messages.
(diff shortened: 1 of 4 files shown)


pages/installation/os/centos.rst
```diff
@@ -92,6 +92,7 @@ The last step is to enable Graylog during the operating system's startup::
   $ sudo systemctl enable graylog-server.service
   $ sudo systemctl start graylog-server.service
 
+The next step is it :ref:`ingest messages <ingest_data>` into your Graylog and extract the messages with :ref:`extractors <extractors>` or use :ref:`the Pipelines <pipelinestoc>` to work with the messages.
```
bernd: I think this should be "The next step is to ingest messages into Graylog and ...".

Same in the other files.

bernd (changes requested):

pages/installation/os/debian.rst
```diff
@@ -13,7 +13,7 @@ Prerequisites
 Not all required dependencies are available in the standard repository, so we need to add `Debian Backports <https://backports.debian.org>`__ to the list of package sources::
 
   $ sudo echo "deb http://ftp.debian.org/debian jessie-backports main" > /etc/apt/sources.list.d/backports.list
-  $ sudo apt-get update 
+  $ sudo apt-get update && apt-get upgrade
```
joschi: Either remove `sudo` from the first command or add it to the second command, but as it is, it's inconsistent.

bernd (approved):

Issue #213, opened 2016-10-20
Add section about using authbind


joschi:
People regularly struggle with the concept of [privileged ports](https://www.w3.org/Daemon/User/Installation/PrivilegedPorts.html) and how to allow Graylog to bind to these (e. g. port 514 for syslog).
- https://en.wikipedia.org/wiki/Authbind
- https://debian-administration.org/article/386/Running_network_services_as_a_non-root_user.
- https://github.com/Graylog2/fpm-recipes/blob/2.0/recipes/graylog-server/files/environment#L10-L12

bernd:
Using `setcap` could be an alternative, although it's not as nice because you have to allow all Java programs to bind to privileged ports.

https://blogs.oracle.com/sduloutr/entry/binding_a_server_to_privileged

bubba198:
I did read the technical explanation and it does not sound credible; if this was a java issue binding to lower ports how does it bind to 514 then?

ok after some tinkering I discovered that the OVA which ships with 2.1.0+62db7e0, codename Smuttynose uses authbind to make it work on 514 so that's how they get it going. Hence I only had to add UDP 162 to the authbind-by-port config and that allowed me to run the snmp plug-in as native 162 listener. Remember to use "!" when specifying a UDP port for authbind-by-port:

touch '/etc/authbind/byport/!162'
chmod 500 '/etc/authbind/byport/!162'
chown graylog '/etc/authbind/byport/!162'

Remember to surround the path with '' because of the ! used in front of the numeric port for UDP. If you do NOT use ! then the port will be TCP and snmp plug-in won't work since snmp is UDP to begin with. Hope that help others to get their plug-in going for native snmp.

```
ubuntu@graylog:/etc/authbind/byport$ ls -la
total 8
drwxr-xr-x 2 root    root    4096 Oct 26 15:33 .
drwxr-xr-x 5 root    root    4096 Oct 26 07:22 ..
-r-x------ 1 graylog graylog    0 Oct 26 15:19 !162
-r-xr-x--- 1 graylog graylog    0 Sep 21 13:35 !514
ubuntu@graylog:/etc/authbind/byport$

```

joschi:
@bubba198 Please don't hijack this issue. Privileged ports are not a "Java issue" but simply restriction of the network stack of the operating system.

You can read up on privileged ports at https://www.w3.org/Daemon/User/Installation/PrivilegedPorts.html

joschi:
[...]

PR #232, opened 2016-12-12
added authbind to faq

jalogisch:
as we do not have any information in our documentation for authbind or how to run inputs on unpriviliges ports we should have at least a small comment in the FAQ.

The final goal would be to have a page in the configuration area that explains how to use this all and how to configure the different options that can be found in: https://github.com/Graylog2/documentation/issues/213

### DIFF
--- a/pages/faq.rst
+++ b/pages/faq.rst
@@ -138,6 +138,16 @@ On Systems that are installed with :ref:`DEB / APT <operationg_package_DEB-APT>`
 
 Systems that are installed with :ref:`RPM / YUM / DNF <operating_package_rpm-yum-dnf>` the file is found in ``/etc/sysconfig/graylog-server``. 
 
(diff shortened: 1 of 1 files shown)


pages/faq.rst
```diff
@@ -138,6 +138,16 @@ On Systems that are installed with :ref:`DEB / APT <operationg_package_DEB-APT>`
 
 Systems that are installed with :ref:`RPM / YUM / DNF <operating_package_rpm-yum-dnf>` the file is found in ``/etc/sysconfig/graylog-server``. 
 
+How can I start my Syslog Input on Port 514?
```
joschi: Maybe generalize this to "a port below 1024".

joschi:
The build currently fails: https://travis-ci.org/Graylog2/documentation/builds/183518271#L158-L159

PR #234, opened 2016-12-19
Multi Server Information to Step-by-Step Guides

jalogisch:
As I have the feeling people do not know how to step from single server step-by-step guide to multi server setups i just had added some additional text about that and also create some refs to the already present documentation. 

I tried to be very generic to cover #138 but also other possible cases.

CLOSE https://github.com/Graylog2/documentation/issues/138

PR #236, opened 2016-12-23
dedub text

jalogisch:
After talked to @bernd about https://github.com/Graylog2/documentation/pull/234 we found the solution to add just a link to the multi-node page and add some more configuration context at this location better. So we do not need to have the same text on different pages.


bernd:
LGTM :+1:

PR #237, opened 2016-12-23
worked on vagrant

jalogisch:
as I was looking for the installation parts I just notice that we have doubled vagrant content.

so I did some modification to that part.


mariussturm:
LGTM!

PR #241, opened 2017-01-12
mention how to fix result window size

kroepke:
("cannot pagination beyond page 66")

fixes Graylog2/graylog2-server#2917


kroepke:
This also applies to 2.1 and should probably be cherry-picked

jalogisch (approved):
👍

PR #254, opened 2017-01-20
Bump version to 2.2.0-beta.6

dennisoelkers:


conf.py
```diff
@@ -65,7 +65,7 @@
 # The short X.Y version.
 version = '2.2'
 # The full version, including alpha/beta/rc tags.
-release = '2.2.0-beta.5'
+release = '2.2.0-beta.6'
```
joschi, line 68: I think we can generally simply use `2.2.0` here instead of updating the documentation for every pre-release.

joschi (changes requested):

conf.py
dennisoelkers: Agreed!

PR #262, opened 2017-01-30
Updates pipelines documentation

edmundoa:
Update pipelines documentation with changes in 2.2:

- Pipelines are not considered experimental any longer, so we remove the warnings
- Add new functions to the functions list
- Update documentation regarding the default stream
- Update usage documentation


pages/pipelines/functions.rst
```diff
@@ -475,11 +477,17 @@ processed for this message.
 
 create_message
 --------------
-``create_message([message: string], [source: string], [timestamp: DateTime])``
+``create_message([message: Message], [source: string], [timestamp: DateTime])``
```
kroepke, line 480: No, this actually is a string, not a message object.

pages/pipelines/functions.rst
```diff
@@ -475,11 +477,17 @@ processed for this message.
...
 
 Creates a new message with from the given parameters. If any of them is omitted, its value is taken from the currently
 processed message. If ``timestamp`` is omitted, the timestamp of the created message will be the timestamp at that moment.
 
+clone_message
+-------------
+``clone_message([message: Message])``
+
+Clones a message. If ``message`` is omitted, this function uses the currently processed message.
```
kroepke: see above it's a string, in case when it's omitted the `message` field of the currently processed message is used.

kroepke (commented):
i'll submit a new PR for the changes

PR #279, opened 2017-02-08
Add description of pattern for parse_date() function

joschi:
The description of the `parse_date()` function lacks a formal description of the pattern string and doesn't even show examples, so that users are literally forced to read the source code or use trial and error.

This PR adds a hint to Joda-Time and adds a description of the pattern.


joschi:
This should also be back-ported to the `2.1` and `2.0` branches.

joschi:
Back-ported in 042050c00fefadd46a4467a384025d5cd56428b9 and bde12f951cd50abc073d6df06f6d9ce935a61ea0.

PR #280, opened 2017-02-15
Add hint to `yum clean all` in RPM instructions

joschi:
Users have been irritated that YUM is in some cases unable to find the latest Graylog package, although the correct repository was being used.

As it turns out, YUM likes using a stale cache and needs a little kick in the cache with `yum clean all`.


joschi:
This should also be back-ported into `2.1` and `2.0`.

jalogisch:
I have back ported to 2.1 this commit - but in 2.0 we do not had the update information in the documentation.

PR #283, opened 2017-02-23
Remove "no duplication" language from streams.rst

dthvt:
Remove incorrect statement that streams do not result in duplication.  With addition of index sets, duplication is possible, as noted earlier in this same page.


joschi:
@dthvt Good catch, thanks!

joschi (approved):

PR #285, opened 2017-03-02
added ports

jalogisch:
to have the default ports written down on the securiy side - to be more clear on that I added them together with some references for SELinux because that is for the user more the point to enter this topic.

### DIFF
--- a/pages/securing.rst
+++ b/pages/securing.rst
@@ -23,7 +23,9 @@ All parts of one Graylog installation will communicate over network sockets. Dep
       - Port
     * - Graylog (web interface / API)
       - 9000 (tcp)
-    * - Elasticsearch (binary protocol)
+    * - Graylog to Elasticsearch
+      - 9350 (tcp)
+    * - Elasticsearch node communication
       - 9300 (tcp)
     * - MongoDB
       - 27017 (tcp)
--- a/pages/securing.rst
+++ b/pages/securing.rst
@@ -19,8 +19,8 @@ All parts of one Graylog installation will communicate over network sockets. Dep
 .. list-table:: Default network communication ports
     :header-rows: 1
 
-    * - Port
-      - Service
+    * - Component
+      - Port
     * - Graylog (web interface / API)
       - 9000 (tcp)
     * - Elasticsearch (binary protocol)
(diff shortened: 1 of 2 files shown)


pages/securing.rst
```diff
@@ -11,6 +11,13 @@ Never run this in an insecure network.
 
 When using Amazon Web Services and our pre-configured AMI, never open all ports in the security group. Do not expose the server to the internet. Access Graylog only from within your VPC. Enable encryption for the communication.
 
+Default ports
+=============
+
+All parts of one Graylog installation will communicate over network sockets. Depending on your setup and number of nodes this might be exposed or can be bound to localhost. The :ref:`SELinux <selinux>` configuration is already covered in our step-by-step guide for centos. The important default network ports to cover are ``9000`` for Graylog, ``9200`` for Elasticsearch and ``27017`` for MongoDB.
```
joschi: We're using the Elasticsearch binary transport protocol (9300/tcp).

joschi: Maybe it would make sense to list the ports and their description in a table for quick lookup.

pages/securing.rst
```diff
@@ -11,6 +11,13 @@ Never run this in an insecure network.
...
+
+Every setup is unique in the requirements and ports might be changed by configuration, but you should limit who is able to connect to what service. In the :ref:`architecture description <big_production_setup>` you can see what parts need to be exposed and communicate with each other.
```
joschi: Every &rarr; Each
what service &rarr; which service
parts &rarr; ports (or "components", depending on what you mean)

joschi (changes requested):

pages/securing.rst
joschi: centos &rarr; CentOS Linux

pages/securing.rst
```diff
@@ -11,6 +11,27 @@ Never run this in an insecure network.
...
+
+.. list-table:: Default network communication ports
+    :header-rows: 1
+
+    * - Component
+      - Port
+    * - Graylog (web interface / API)
+      - 9000 (tcp)
+    * - Elasticsearch (binary protocol)
+      - 9300 (tcp)
```
joschi: Graylog itself is using port 9350/tcp for communicating with Elasticsearch (which is using 9300/tcp by default).

https://github.com/Graylog2/graylog2-server/blob/2.2.1/misc/graylog.conf#L291-L292

joschi (changes requested):

joschi (approved):

PR #286, opened 2017-03-03
added faq for azure install

jalogisch:
just to have something cover how to run and install graylog on azure.


dthvt:
I think you should remove the word "just" in the added text.  i.e.

You can `create a Linux VM...

As a second option you can use `this guide...

I think it reads more professionally.

joschi (approved):

PR #294, opened 2017-03-14
worked on alerts

jalogisch:
I just updated the alerts page to have written down what is able in GL 2.2.2 - just to have repeated notification enabled


edmundoa (approved):
LGTM 👍

PR #296, opened 2017-03-28
Fix a typo

etlweather:
There was a typo in the heading


etlweather:
There is another pull request already for that fix.

PR #299, opened 2017-04-05
Update archive documentation

bernd:
Updates the archive documentation to 2.2.3 features.

- Replace `archive_path` setting with backends
- Describe output path template for backends
- Add missing checksum type parts
- Some adjustments wrt index sets
- Update all screenshots to show the current UI


jalogisch (changes requested):
It should be mentioned that the base path for the archives should be created and writable to graylog. 

If you use variable the path until the first variable should be already present.

bernd:
> It should be mentioned that the base path for the archives should be created and writable to graylog.

I clarified the sentence about accessing the directory.

jalogisch (approved):
LGTM

PR #303, opened 2017-04-20
Catch up with enterprise releases

mariussturm:


pages/enterprise/changelog.rst
```diff
@@ -2,6 +2,63 @@
...
 
+Graylog Enterprise 2.2.3
+========================
+
+Released: 2017-04-04
+
+Plugin: Archive
+---------------
+
+Metadata is now stored in MongoDB
```
joschi: These should probably be bullet lists, so that each item is shown on its own line.

http://www.sphinx-doc.org/en/stable/rest.html#lists-and-quote-like-blocks

joschi (changes requested):

pages/enterprise/changelog.rst
mariussturm: True, didn't saw that it's one-line otherwise.

pages/enterprise/changelog.rst
```diff
@@ -2,6 +2,63 @@
...
+Plugin: Archive
+---------------
+
+Prepare for the plugin to be compatible with the new default stream.
+
+Plugin: Audit Log
+-----------------
+
+Add support for index sets and fix potential NPEs.
+Smaller UI imprevements.
```
joschi: Same here. Should be a bullet list so that these are printed on their own lines each.

joschi: In the end, all change items should be in a bullet list, for a consolidated look.

joschi (approved):

PR #307, opened 2017-05-12
hint that login and example are intentionally misspelled

jsoref:
Proposal based on #306


joschi:
@jsoref Thanks for your contribution!

I don't think that this change improves the reading flow or the semantics of the sentence, so I'll close this one without merging.

jsoref:
So. I'm open to other suggestions for hinting that the following items are intentionally misspelled, but it certainly would imo help to highlight that the tokens below are misspelled.

Ignoring the misspelled portion, the original text is incredibly stilted/awkward.
"not equal terms" doesn't really make sense...
(At this point, I understand what you're trying to say, but the current text isn't proper...)

PR #324, opened 2017-06-30
Document lookup tables

bernd:


pages/lookuptables.rst
```diff
@@ -0,0 +1,129 @@
...
+
+**Example 2:** Output for the geo-ip data adapter including a single value
+and a multi value.
+
+.. image:: /images/lookuptables/multi-value-example.png
+
+Setup
+-----
+
+The lookup tables can be configured on the "System/Lookup Tables" page.
```
jalogisch, line 80: in other parts of the Docs we use `System > Lookup Tables` (Code Markup) that should be identical in all parts

jalogisch (commented):
I would change that before commit - but please comment.

pages/lookuptables.rst
bernd: We are mostly using the ` / ` notation:

```
$ git grep -e 'System >' -e 'System>' | wc -l
3
$ git grep -e 'System /' -e 'System/' | wc -l
29
```

In some places we also use ` -> `. Might be nice to unify that at one point.

jalogisch (approved):
here we go

pages/lookuptables.rst
jalogisch: i'll cover that

PR #328, opened 2017-07-14
clarify csv  cache

jalogisch:
because of https://github.com/Graylog2/graylog2-server/issues/4001 i think we need to explain that behavior in the documentation


pages/lookuptables.rst
```diff
@@ -39,6 +39,11 @@ and do not have to implement it on their own.
 
 Cache implementations are pluggable and new ones can be added through plugins.
 
+.. note:: The CSV file adapter reads the entire file into memory and re-read on
```
joschi: > The CSV file adapter reads the entire file into memory and refreshes its contents after within each check interval if the file was changed.

pages/lookuptables.rst
```diff
@@ -39,6 +39,11 @@ and do not have to implement it on their own.
...
+         the check interval if something is changed. If the cache is purged but the
```
joschi: > If the cache was purged but the check interval has not elapsed, lookups might return expired values.

joschi (changes requested):

pages/lookuptables.rst
```diff
@@ -39,6 +39,11 @@ and do not have to implement it on their own.
 
 Cache implementations are pluggable and new ones can be added through plugins.
 
+.. note:: The CSV file adapter reads the entire file into memory and refreshes
+          its contents after within each check interval if the file was
```
joschi: "after within" &rarr; "after"

joschi (changes requested):

joschi (approved):

PR #346, opened 2017-08-10
Fix for docker run elasticsearch

lodeblomme:
I tried the command with "-Des.cluster.name" and received an error that -D isn't a recognised option.

According to https://www.elastic.co/guide/en/elasticsearch/reference/current/docker.html, it should be "-Ecluster.name". I tested it and it worked.

"-Ees.cluster.name" also gives an error:
org.elasticsearch.bootstrap.StartupException: java.lang.IllegalArgumentException: unknown setting [es.cluster.name] did you mean [cluster.name]?


joschi:
Duplicate of #335.

joschi:
@lodeblomme Your comments on #348 would be much appreciated!

PR #347, opened 2017-08-10
Add instructions for non-root operation

mariussturm:


pages/collector_sidecar.rst
```diff
@@ -367,6 +367,42 @@ Certificate based client authentication
...
+  User=collector
+  Group=collector
+
+To make use of these settings reload systemd::
+
+  $ sudo systemctl daemon-reload
+  $ sudo systemctl restart collector-sidecar
+
+Check the log files in ``/var/log/graylog/collector-sidecar`` for any errors. Understand that not only the Sidecar but also all backends, like ``filebeat``, will be started as ``collector`` user after these changes.
+So all log files that the backound should observe also need to be readable by the ``collector`` user.
```
joschi: Typo: "backend"

joschi: Given this restriction, I think we should add a few sentences about adding the `collector` user to a group which can read most logs.

For example on Debian and Ubuntu Linux, this would be the `adm` group.

pages/collector_sidecar.rst
```diff
@@ -367,6 +367,42 @@ Certificate based client authentication
...
+  - ``cache_path`` - /var/cache/graylog/collector-sidecar/
+  - ``log_path`` - /var/log/graylog/collector-sidecar/
+
+So to make these directories readable for the ``collector`` user, use:
+
+  - ``chown -R collector /etc/graylog``
+  - ``chown -R collector /var/cache/graylog``
+  - ``chown -R collector /etc/graylog``
+
+You can change all pathes to different places in the filesystem. If you prefer to store all Sidecar data in the home directory of the ``collector`` user, just change the pathes accordingly.
```
joschi: Typo: "paths"

joschi (changes requested):

pages/collector_sidecar.rst
```diff
@@ -367,6 +367,43 @@ Certificate based client authentication
...
+  Group=collector
+
+To make use of these settings reload systemd::
+
+  $ sudo systemctl daemon-reload
+  $ sudo systemctl restart collector-sidecar
+
+Check the log files in ``/var/log/graylog/collector-sidecar`` for any errors. Understand that not only the Sidecar but also all backends, like ``filebeat``, will be started as ``collector`` user after these changes.
+So all log files that the backend should observe also need to be readable by the ``collector`` user. Depending on the Linux distribution there is usually an adminstrator group which has access to most log files.
+By adding the `collector` user to that group you can grant access fairly easy. E.g. on Debian/Ubuntu systems this group is called `adm`.
```
joschi: The single back-ticks should be double back-ticks to print a "pre-formatted" string in RST (as compared to Markdown). The single back-ticks will print italic text.

joschi: Sentences shouldn't start with "E. g.", using "For example, " in this case sounds better in my opinion.

joschi (changes requested):

joschi (approved):

PR #348, opened 2017-08-10
Update and rework the Docker installation chapter

joschi:
Closes #335
Closes #343
Closes #346
Closes #349
Closes #350
Closes #351

PR #351, opened 2017-08-11
Update docker.rst

RokPapez:
Wrong elasticsearch option:
When running command written above I get: ERROR: D is not a recognized option. To configure elasticsearch settings we should use:
-E <KeyValuePair>     Configure a setting.


joschi:
Duplicates #348.

PR #361, opened 2017-08-30
Add subjectAltName to self-signed certificates

### LINKED ISSUE #360: HTTPS/SSL - Change description to Subject Alternative Name
## Context
the usage of [common name is deprecated](https://tools.ietf.org/html/rfc2818#page-5) and we should switch the how-to to use subject alternative name.

http://pro-tips-dot-com.tumblr.com/post/65411476159/self-signed-ssl-certificates-with-multiple-hostnames

joschi:
Closes #360


jalogisch (approved):
LGTM

PR #362, opened 2017-09-07
Fix order of parameters in OpenSSL command

joschi:
The `-config` parameter was incorrectly placed between `-newkey` and the value `rsa:2048` which caused the command to not work.


jalogisch (approved):
LGTM

PR #363, opened 2017-09-19
Add description of prefix/suffix parameters of set_field/set_fields

joschi:
Refs Graylog2/graylog-plugin-pipeline-processor/pull/75


jalogisch (approved):
LGTM - good catch on that

PR #387, opened 2017-11-15
Add documentation for the debug() function

joschi:
Refs Graylog2/graylog-plugin-pipeline-processor#188

### DIFF
--- a/pages/pipelines/functions.rst
+++ b/pages/pipelines/functions.rst
@@ -229,6 +229,12 @@ Print any passed value as string in the Graylog log.
 
 Note that the debug message will only appear in the log of the Graylog node that was processing the message you are trying to debug.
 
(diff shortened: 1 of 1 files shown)


joschi:
This PR has to be merged into the `2.4` and `3.0` branches, too.

jalogisch:
shouldn't we add on example? 

Cause something like: ` debug( concat("dropped message from ", to_string($message.source)) );` is not very easy to reach without deeper knowledge.

Or would you see that in a FAQ?

joschi:
@jalogisch I left out the example for consistency, as no other function has an example with it.

This being said, I think you're right and users will profit from a specific example in this case.

jalogisch (approved):
LGTM

Issue #384, opened 2017-11-06
Hunting for graylog-ctl script to find the upgrade notes is not intuitive.


123dev:
A Graylog user had wrongly applied the incorrect update.
That was discussed in [here](https://community.graylog.org/t/server-dead-after-2-3-1-to-2-3-2-upgrade/2935)

I commented the [following](https://community.graylog.org/t/server-dead-after-2-3-1-to-2-3-2-upgrade/2935/5?u=123dev)

> I would say that it is an easy mistake to make considering that both packages are named graylog_latest.deb, they’re just downloaded from different endpoints.
>
> Of course this in context of upgrades.
> Speaking of which, I personally think that upgrade notes should be rooted at one common document.
> Example: Upgrading to 2.3.2, and in that document either describe in sections upgrade of each type, or just link to them.
>
> Hunting for graylog-ctl script to find the upgrade notes is not intuitive.
>
> For all the upgrades I’ve gone through I always thought about suggesting this, but I stopped myself short thinking perhaps it’s just me, after-all I wish other products had as thorough documentation as Graylog.
> 
> Thanks for great product and documentation that goes along with it.

Jochen [agreed](https://community.graylog.org/t/server-dead-after-2-3-1-to-2-3-2-upgrade/2935/6?u=123dev) and asked for a ticket to be logged here.

Thanks

PR #395, opened 2017-12-13
add more Info on Upgrade

jalogisch:
I notice that in the past month several ppl could not identify the way how to update Graylog. This little added text should give them the guidance they need.

FIX: https://github.com/Graylog2/documentation/issues/384

### DIFF
--- a/pages/upgrade.rst
+++ b/pages/upgrade.rst
@@ -2,11 +2,11 @@
 Upgrading Graylog
 *****************
 
-When upgrading from a previous version of Graylog you basically follow the previous used installation method using the new version numbers. 
+When upgrading from a previous version of Graylog you follow the previous used installation method using the new version numbers. 
 
 The following Upgrade notes should be read carefully before you start the upgrade process. Breaking changes and dependency upgrades are documented in those upgrade notes.
 
-You should always follow the version trail and not jump a few versions up leaving out one or more versions. The upgrade notes are always written coming from the stable release before.
+You should always follow minor versions when updating across multiple versions to make sure necessary migrations are run corretly. The upgrade notes are always written coming from the stable release before.
 
 .. toctree::
    :titlesonly:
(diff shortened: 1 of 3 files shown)


jalogisch:
this need to be cherry-picked into 2.4 and 3.0 once merged

pages/upgrade.rst
```diff
@@ -2,8 +2,19 @@
 Upgrading Graylog
 *****************
 
+When upgrading from a previous version of Graylog you basically follow the previous used installation method using the new version numbers.
```
kroepke: "...you follow the previously used installation method..."

pages/upgrade.rst
```diff
@@ -2,8 +2,19 @@
 Upgrading Graylog
 *****************
 
+When upgrading from a previous version of Graylog you basically follow the previous used installation method using the new version numbers. 
+
+The following Upgrade notes should be read carefully before you start the upgrade process. Breaking changes and dependency upgrades are documented in those upgrade notes.
+
+You should always follow the version trail and not jump a few versions up leaving out one or more versions. The upgrade notes are always written coming from the stable release before.
```
kroepke: I'd write:
> You should always follow minor versions when updating across multiple versions to make sure necessary migrations are run correctly.

pages/upgrade.rst
```diff
@@ -2,8 +2,19 @@
...
 .. toctree::
    :titlesonly:
    :glob:
 
    upgrade/graylog-*
+
+
+The Virtual Machine Appliance (OVA) and Amazon Web Services (AMI) Version of Graylog use the Omnibus package. The upgrade documentation is :ref:`part of the graylog-ctl documentation <upgrade_graylog_omnibus>`.
+
+Should the previous installation include the Graylog repository of the current Graylog version, update the repository package to the target version and use the system tools to upgrade the package. For .rpm based systems :ref:`this update guide <operating_package_upgrade_rpm-yum-dnf>` and for .deb based systems :ref:`this update guide <operating_package_upgrade_DEB-APT>` should help.
```
kroepke: I would simplify this a bit:

> Should the current installation use operating system packages, update the repository package to the target version and use the system tools to upgrade the package.

kroepke (changes requested):

pages/upgrade.rst
jalogisch: I had that before in such a simple way and notice that some users did not know how to do. Additional I want to link from that page to the actual page where the update is described.

kroepke (approved):

PR #402, opened 2018-01-10
Update installation.rst

unamuno:
ElasticSeach 6 is not suported yet. But the docu does not tell you.
https://community.graylog.org/t/deflector-exists-as-an-index-and-is-not-an-alias-reopen/3463/3


joschi (approved):
LGTM. Thanks!

PR #406, opened 2018-01-20
Update doc to give more verbose information

211217613:
When getting started with the vagrant file, it's not clear where or how to generate or copy the server.conf file from or too.

### DIFF
--- a/pages/configuration/server.conf.rst
+++ b/pages/configuration/server.conf.rst
@@ -2,7 +2,7 @@
 server.conf
 ***********
 
-The file ``server.conf`` is the Graylog configuration file.
+The file ``server.conf`` is the Graylog configuration file. It is not auto generated. You will need to copy this file to `/etc/graylog/server/` directory that will need to be created.
 
 .. note:: Check :ref:`default_file_location` to locate it in your installation.
 


joschi:
@211217613 Pretty much all installation methods for Graylog come with a default configuration file (also see http://docs.graylog.org/en/2.4/pages/configuration/file_location.html).

I haven't checked if this is different for the Vagrant box, but if it is, it might be better to add this information to the README file in the repository hosting the Vagrant box:
https://github.com/Graylog2/graylog2-images/tree/2.4/vagrant

211217613:
>  come with a default configuration file

I'm not sure if this is obvious to others or I'm just the only one having trouble with it. All I'm trying to do is setup HTTPS  for rest and web interface. The HTTPS page seems somewhat straightforward but when setting up a test environment with the vagrantfile, their is no sample config file in `/etc/graylog/server/server.conf`. There is a sample config file in `/opt/graylog/server/graylog.conf.example` which is inconsistent with the documentation. The documentation reads like their should be a default `server.conf` in `/etc/graylog/server/server.conf` and also another configuration file in `/opt/graylog/conf/graylog.conf`

PR #420, opened 2018-02-22
Update MongoDB install section

jasonkeller:
Updated MongoDB install section to reflect using the latest official package on Ubuntu from the MongoDB maintainers, rather than point people to an ancient version that is prohibitive to upgrade.

### DIFF
--- a/pages/installation/os/debian.rst
+++ b/pages/installation/os/debian.rst
@@ -19,9 +19,12 @@ If you're starting from a minimal server setup, you will need to install these a
 MongoDB
 -------
 
-The version of MongoDB included in Debian Jessie is recent enough to be used with Graylog 2.3.x and higher::
+The official MongoDB repository provides the most up-to-date version and is the recommended way of installing MongoDB::
 
-  $ sudo apt install mongodb-server
+  $ sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 58712A2291FA4AD5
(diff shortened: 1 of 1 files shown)


pages/installation/os/debian.rst
```diff
@@ -19,9 +19,12 @@ If you're starting from a minimal server setup, you will need to install these a
 MongoDB
 -------
 
-The version of MongoDB included in Debian Jessie is recent enough to be used with Graylog 2.3.x and higher::
+The official MongoDB repository provides the most up-to-date version and is the recommended way of installing MongoDB::
 
-  $ sudo apt install mongodb-server
+  $ sudo apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 58712A2291FA4AD5
+  $ echo "deb http://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/3.6 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-3.6.list
```
jalogisch, line 25: that looks more like a line for ubuntu and not debian. 

according to https://docs.mongodb.com/manual/tutorial/install-mongodb-on-debian/#create-a-etc-apt-sources-list-d-mongodb-org-3-6-list-file-for-mongodb

this is not a valid line for debian and should be changed.

jalogisch (changes requested):
Thank you @jasonkeller for the PR 

Would you mind to look at the comment and maybe include the section about Ubuntu and not just one?

PR #424, opened 2018-02-27
Update lookuptables.rst

jasonkeller:
Updated to make clear that the lookup tables use HEAP memory, and that it must be sized accordingly.


joschi (approved):
LGTM. 👍

PR #428, opened 2018-03-22
Update queries.rst

knechtionscoding:
Fixing two typos of the word example.


joschi:
@KnechtionsCoding Thanks for your contribution!

As you can see from the context of these examples, the typos are there on purpose to demonstrate fuzzy search.

PR #431, opened 2018-04-04
link to releases list

sashaman:
either update to proposed change or prevent the 404 of actual link


joschi:
Fixed via 8d89c0d222e04d5a47979074bbbf4f2c46d92191.

PR #432, opened 2018-04-04
link to releases

sashaman:
currentliy 404's so change to working link, or if preferred SSL use githubs https://github.com/Graylog2/graylog2-server/releases 
since HTTPS with https://docs.graylog.org/ complains about SSL_ERROR_BAD_CERT_DOMAIN (*.)readthedocs.org


joschi:
Fixed via 8d89c0d222e04d5a47979074bbbf4f2c46d92191.

PR #434, opened 2018-04-11
Fix typo.

ejball:


joschi (approved):
@ejball Good catch. Thank you very much!

PR #435, opened 2018-04-12
added note about keystore location

jalogisch:
added note about that this addition is needed to make it work

was done after customer feedback and several ppl struckle with that.

should be cherrypicked to 2.4/2.3 docs

### DIFF
--- a/pages/configuration/https.rst
+++ b/pages/configuration/https.rst
@@ -281,7 +281,7 @@ Most start and init scripts for Graylog provide a ``JAVA_OPTS`` variable which c
 
 .. note:: The default location to change the JVM parameter depends on your installation type and is documented :ref:`with all other default locations <default_file_location>`.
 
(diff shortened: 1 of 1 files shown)


pages/configuration/https.rst
```diff
@@ -279,6 +279,10 @@ In order for the JVM to pick up the new trust store, it has to be started with t
 
 Most start and init scripts for Graylog provide a ``JAVA_OPTS`` variable which can be used to pass the ``javax.net.ssl.trustStore`` and (optionally) ``javax.net.ssl.trustStorePassword`` system properties.
 
+.. note:: The default location to change the JVM parameter depends on your installation type and is documented :ref:`with all other default locations <default_file_location>`.
+
+.. warning:: Without adding the keystore to your startup parameter Graylog is not able to work with your self-signed certificates or your local CA signed certificates.
```
joschi:
```text
Without adding the previously created Java keystore to the JVM parameters, Graylog won't be able to verify any self-signed certificates or custom CA certificates.
```

joschi (changes requested):

PR #436, opened 2018-04-12
fix tar download

jalogisch:
included the direct download link to the tar ball in the docs

That would need to be adjusted/added to 3.0 branch too


jalogisch (commented):
<3

joschi (approved):

PR #440, opened 2018-04-26
Update debian.rst

kurobeats:
Corrected repo list as per https://docs.mongodb.com/manual/tutorial/install-mongodb-on-debian/


joschi:
CC: @jalogisch 

This was just changed in 142b19f5c8bec6475140fd19e013e719dfd3171e and I tend to use the value from the official MongoDB documentation. 😉

PR #447, opened 2018-06-11
Add RegExp query doc

steelywing:


pages/queries.rst
```diff
@@ -45,6 +45,10 @@ Messages that do not have the field *type*::
 
   NOT _exists_:type
 
+Messages that match RegExp *FastEthernet[0-9]+* (ElasticSearch will convert messages to lowercase)::
```
joschi: Please write "regular expression" instead of "RegExp".

joschi: "ElasticSearch" -> "Elasticsearch"

joschi: In fact, Elasticsearch will only convert the query to lower case, if the configured tokenizer does that:
https://www.elastic.co/guide/en/elasticsearch/reference/5.6/analysis-lowercase-tokenizer.html

For example, the standard analyzer of Elasticsearch is using the lowercase tokenizer:
https://www.elastic.co/guide/en/elasticsearch/reference/5.6/analysis-standard-analyzer.html

pages/queries.rst
```diff
@@ -45,6 +45,10 @@ Messages that do not have the field *type*::
...
+
+  /fastethernet[0-9]+/
```
joschi: Either change to "/FasterEthernet[0-9]+/" here or to "fasterethernet[0-9]+" in the above paragraph.

pages/queries.rst
joschi: Please use back-quotes to emphasize the regular expression:
http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup

```text
``FastEthernet[0-9]+``
```

joschi (changes requested):

pages/queries.rst
steelywing: But if I search `FastEthernet[0-9]+`, the result is empty, if I search `fastethernet[0-9]+`, the result is normal, may be I just use all lowercase in this example?

joschi: As explained, it depends on the tokenizer which is being used for the respective field(s).

Maybe add examples and an explanation for when lowercase searches work and when not.

steelywing: Actually I don't know what is tokenizer? 😄 I just know the lowercase works...

steelywing: Can you help to add the explaination? 😃 @joschi

joschi (dismissed):

Issue #446, opened 2018-06-08
Collector Sidecar first start clarification


20k-ultra:
As of Debian 9 it uses systemd so I suggest updating the docs [here](http://docs.graylog.org/en/2.4/pages/collector_sidecar.html#first-start) so that Debian is with RedHat/CentOS on how to start the service. I think this is appropriate since Debian 9 is the latest version.

* Graylog Version: 2.4
* Operating System: Debian 9

joschi:
@20k-ultra Would you like to contribute a pull request to the documentation changing this aspect?

It would be great to still have some explanation for Upstart/SysV init, too.

20k-ultra:
sure I'll give it a try!

PR #451, opened 2018-07-18
modified first start table to be initsystem based

20k-ultra:
PR for bug #446. I think since upstart was only used in Ubuntu which has since switched to SystemD since 2015 then it's pretty safe to assume most users aren't using it anymore. Also mostly due to the fact that a .conf file will be required to setup the upstart process so it's not as simple as the other initsystems.

However, if anyone thinks it should be included then I'll add it. I would just ask someone let me know how I should go about adding it to the project. Create instructions below the table on how to create a upstart .conf file for Graylog ?

Lastly, I know there's a trade off to listing initsystems vs distros (before it was distros) but I think this is much better to maintain since some distros recently switched so someone running older version of a distro could be running an initsystem that isn't specified in the table.

note: this is my first PR so if I've done anything wrong please let me know. I was unable to find a contributing portion in the readme which perhaps would have helped a new contributors.

### DIFF
--- a/pages/collector_sidecar.rst
+++ b/pages/collector_sidecar.rst
@@ -280,13 +280,19 @@ First start
 Once you installed the Sidecar package you are ready to start the service for the first time. Decide which backend you want to use. Enable or disable the single
 backends by setting ``enabled: true`` or respectively to ``false``. Now start the Sidecar, depending on your operating system you can do this with:
 
-+---------------+---------------------------------------------------------------------------------------------+
-| Debian/Ubuntu | ``sudo start collector-sidecar``                                                            |
-+---------------+---------------------------------------------------------------------------------------------+
-| RedHat/CentOS | ``sudo systemctl start collector-sidecar``                                                  |
-+---------------+---------------------------------------------------------------------------------------------+
-| Windows       | ``C:\Program Files\graylog\collector-sidecar\graylog-collector-sidecar.exe -service start`` |
-+---------------+---------------------------------------------------------------------------------------------+
++-------------+----------------------------------------------------------------------------------------------+
+| systemd     | ``sudo systemctl start collector-sidecar``                                                   |
++-------------+----------------------------------------------------------------------------------------------+
+| SysV        | ``sudo start collector-sidecar``                                                             |
++-------------+----------------------------------------------------------------------------------------------+
+| Windows     | ``C:\Program Files\graylog\collector-sidecar\graylog-collector-sidecar.exe -service start``  |
++-------------+----------------------------------------------------------------------------------------------+
+
+If you're unsure which init system your Linux distribution is using, execute the following command to print the name of the used init system::
+
+    # ps -h -o comm -p 1
+
+Otherwise please refer to the handbook of your Linux distribution and look up which init system is being used.
 
 Afterwards you will most likely see an error like this in the log file::
 


pages/collector_sidecar.rst
```diff
@@ -280,13 +280,19 @@ First start
...
 
-+---------------+---------------------------------------------------------------------------------------------+
-| Debian/Ubuntu | ``sudo start collector-sidecar``                                                            |
-+---------------+---------------------------------------------------------------------------------------------+
-| RedHat/CentOS | ``sudo systemctl start collector-sidecar``                                                  |
-+---------------+---------------------------------------------------------------------------------------------+
-| Windows       | ``C:\Program Files\graylog\collector-sidecar\graylog-collector-sidecar.exe -service start`` |
-+---------------+---------------------------------------------------------------------------------------------+
++-------------+----------------------------------------------------------------------------------------------+
+| SystemD     | ``sudo systemctl start collector-sidecar``                                                   |
```
joschi: The name is ["systemd"](https://www.freedesktop.org/wiki/Software/systemd/), not "SystemD". 😉

pages/collector_sidecar.rst
```diff
@@ -280,13 +280,19 @@ First start
...
++-------------+----------------------------------------------------------------------------------------------+
+| SysV        | ``sudo start collector-sidecar``                                                             |
++-------------+----------------------------------------------------------------------------------------------+
+| Windows     | ``C:\Program Files\graylog\collector-sidecar\graylog-collector-sidecar.exe -service start``  |
++-------------+----------------------------------------------------------------------------------------------+
+
+If you're unsure which initsystem your distro uses try this one liner::
+
+    for initsys in systemd upstart init; do ps x | grep -qi "$initsys" && echo $initsys && break; done
```
joschi: I'd rather use the following command, which shows the command of PID 1:
```text
# ps -h -o comm -p 1
```

pages/collector_sidecar.rst
```diff
@@ -280,13 +280,19 @@ First start
...
-+---------------+---------------------------------------------------------------------------------------------+
++-------------+----------------------------------------------------------------------------------------------+
+| SystemD     | ``sudo systemctl start collector-sidecar``                                                   |
++-------------+----------------------------------------------------------------------------------------------+
+| SysV        | ``sudo start collector-sidecar``                                                             |
++-------------+----------------------------------------------------------------------------------------------+
+| Windows     | ``C:\Program Files\graylog\collector-sidecar\graylog-collector-sidecar.exe -service start``  |
++-------------+----------------------------------------------------------------------------------------------+
+
+If you're unsure which initsystem your distro uses try this one liner::
```
joschi: Please change to:
```text
If you're unsure which init system your Linux distribution is using, execute the following command to print the name of the used init system::
```

pages/collector_sidecar.rst
```diff
@@ -280,13 +280,19 @@ First start
...
+
+    for initsys in systemd upstart init; do ps x | grep -qi "$initsys" && echo $initsys && break; done
+
+Otherwise look up your distro and version since some distros could have switched their initsystem over time.
```
joschi: Please change to:
```text
Otherwise please refer to the handbook of your Linux distribution and look up which init system is being used.
```

joschi (changes requested):
@20k-ultra Thanks for your contribution!

Please sign the CLA for this repository at https://cla-assistant.io/Graylog2/documentation?pullRequest=451 and address the requested changes.

pages/collector_sidecar.rst
20k-ultra: agreed. much cleaner.

20k-ultra:
I didn't have my email set on my git config so when I made the commit github is unable to identify the first commit as my own. Now that I set my email and made another commit it's working however  I the CLA says the the previous commit from Miguel Casqueira (me) has not signed but I have. Any input on how to fix that commit ? I think I have to find what the email is used for that commit and add it to my github profile.

edit: decided to squash the commits since now the email issue is fixed and squashing will fix remove that broken account.

joschi (approved):
Looks good to me. 👍 
Thanks for your contribution!

PR #460, opened 2018-07-26
Update collector sidecar documentation

kyrogue:
The provided example configuration is not valid for a windows configuration

### DIFF
--- a/pages/collector_sidecar.rst
+++ b/pages/collector_sidecar.rst
@@ -250,12 +250,13 @@ For the Beats platform you can enable each Beat individually, e.g on a Windows h
     list_log_files:
       - /var/log
     node_id: graylog-collector-sidecar
-    collector_id: file:/etc/graylog/collector-sidecar/collector-id
-    log_path: /var/log/graylog/collector-sidecar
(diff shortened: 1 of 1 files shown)


kyrogue:
can someone confirm the line 

> collector_id: file:C:\Program Files\graylog\collector-sidecar\collector-id

actually takes has a 'file' value infront of the path?

mariussturm:
Hi,
you need double slashes for `collector_id`. It's in the format like:
```
collector_id: "file:C:\\Program Files\\graylog\\collector-sidecar\\collector-id
```
Additionally the `list_log_files` option need to be changed for windows.

kyrogue:
Running the graylog_collector installer.exe in windows server 2012, generates what i pasted as the default configuration settings. Or do you mean i need to double slash it to escape the slash in markdown?
What could be a reasonable example to display for the list_log_files directory for windows?

mariussturm:
Ah ok, guess both formats are fine then.
Either you set `list_log_files` to an empty list to disable it or to something like `C:\Windows\System32\winevt\Logs` to see activity on the different event channels. But the unix path doesn't make sense on windows.

mariussturm (approved):

PR #464, opened 2018-08-19
Update functions.rst

megan201296:
Add examples that may be useful to new users of pipelines.

### DIFF
--- a/pages/pipelines/functions.rst
+++ b/pages/pipelines/functions.rst
@@ -336,8 +336,8 @@ Checks if ``value`` contains ``search``, optionally ignoring the case of the sea
 
 Example::
 
-        // Check if the `computer.corp.net` is in the `hostname` field. Ignore case.
-        contains(to_string($message.hostname), "computer.corp.net", true)
+        // Check if the `example.org` is in the `hostname` field. Ignore case.
+        contains(to_string($message.hostname), "example.org", true)
 
 substring
 ---------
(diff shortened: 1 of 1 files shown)


pages/pipelines/functions.rst
```diff
@@ -322,6 +334,11 @@ contains
 
 Checks if ``value`` contains ``search``, optionally ignoring the case of the search pattern.
 
+Example::
+
+        // Check if the `computer.corp.net` is in the `hostname` field. Ignore case.
+        contains(to_string($message.hostname), "computer.corp.net", true)
```
jalogisch: could you please change that to `example.org` as we use that all over the documentation.

pages/pipelines/functions.rst
```diff
@@ -508,6 +525,11 @@ processed for this message.
 
 If ``remove_from_default`` is ``true``, the message is also removed from the default stream "All messages".
 
+Example::
+
+        // Route the current processed message to a stream named `Custom Stream`
+        route_to_stream("Custom Stream");
```
jalogisch: using the name is onyl the second best option - could you also add example for the route to a streamid.

From a performance perspective this is prefered and in addition the name can be assigned multiple times. The ID is uniq.

jalogisch (changes requested):
thank you for this little update. We need to improve this part of the documentation and giving some samples is a nice to have.

Could you please check on my comments.

pages/pipelines/functions.rst
megan201296: Done!

pages/pipelines/functions.rst
megan201296: Done. I've never actually used the ID. I believe that it goes in quotes, is that correct? Add it that way but feel free to fix if incorrect.

megan201296:
Made the requested changes @jalogisch! Thanks for the feedback.

jalogisch (approved):

jalogisch:
thank you for the contribution @megan201296

PR #476, opened 2018-09-10
added note about field compare

jalogisch:
after realisation that it is possible to compare two fields I add the note about.

need to be merged to 3.0 too

### DIFF
--- a/pages/pipelines/rules.rst
+++ b/pages/pipelines/rules.rst
@@ -96,6 +96,7 @@ The condition must not be empty, but can simply consist of the boolean literal `
 
 If a condition calls a function which is not present (perhaps due to a typo or missing plugin) the call evaluates to ``false``.
 
+.. note:: Comparing two fields can be done when you use the same data type, e.g.  ``to_string($message.src_ip) == to_string($message.dst_ip)`` will compare the two strings and will become ``true`` on match.
 
 Actions
 =======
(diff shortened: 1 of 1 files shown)


pages/pipelines/rules.rst
```diff
@@ -96,6 +96,7 @@ The condition must not be empty, but can simply consist of the boolean literal `
 
 If a condition calls a function which is not present (perhaps due to a typo or missing plugin) the call evaluates to ``false``.
 
+.. note:: Comparing two fields can be done when you use the same data type, e.g.  ``to_string($message.src_ip) == to_string($message.dst_ip)`` will compare the two strings and will become ``true`` on match.
```
mpfz0r: n00b reviewer question: what happens if you try to compare different data types?

jalogisch: I just tried that and it just fails - because it can't compared.

mpfz0r: is that visible to the user?

jalogisch: the `when` condition did not become true - so `then` is not executed. 

Not directly visible, but the rule just did not work.

mpfz0r: I was wondering if that's noteworthy.  e.g  "Comparing different data types evaluates to ``false``".

jalogisch: yes, good catch!

mpfz0r (approved):

Issue #433, opened 2018-04-06
Add section about regular expressions in search queries


joschi:
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-regexp-query.html#regexp-syntax

PR #477, opened 2018-09-10
remove drools / added exclude to generation

jalogisch:
- added lib folder to be excluded on side generation to avoid warnings local (on mac)

FIX: https://github.com/Graylog2/documentation/issues/433


pages/drools.rst
```diff
@@ -1,124 +1,10 @@
...
-          }
-  end
-
+******
+Drools
+******
 
-Blacklisting messages
-=====================
+Drools support is removed since Graylog 3.0. It can be integrated with a Plugin - but currently no active development happen.
```
bernd: Lets just write "Drools support got removed in Graylog 3.0." without the second sentence. No need to tell people about the plugin if there is none. :wink:

pages/blacklisting.rst
```diff
@@ -1,46 +1,9 @@
+:orphan:
+
 .. _blacklisting:
 
 ************
 Blacklisting
 ************
 
 .. note:: Since Graylog 2.0 you can use the :doc:`processing pipelines <pipelines>` for blacklisting.
```
bernd: Is there a black listing example on the processing pipeline page that we can link to?

bernd (changes requested):

pages/blacklisting.rst
jalogisch: currently not - but I'll add one and then link to that

bernd (approved):

PR #488, opened 2018-10-04
Update functions.rst

megan201296:
Added additional examples of various functions. Also added a reference to the test pipeline rules in the graylog server repo as it is a "not-easy-to-find" but useful resource.

### DIFF
--- a/pages/pipelines/functions.rst
+++ b/pages/pipelines/functions.rst
@@ -79,6 +79,8 @@ In our case the alternative version of calling ``parse_date`` would look like th
         set_field("transaction_year", new_date.year);
     end
 
+More examples of the usage of various functions in pipeline rules can be found in the `graylog2-server repo <https://github.com/Graylog2/graylog2-server/tree/master/graylog2-server/src/test/resources/org/graylog/plugins/pipelineprocessor/functions>`_.
+
 All parameters in Graylog's processing functions, listed below, are named.
 
 Function Index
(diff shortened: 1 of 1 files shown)


pages/pipelines/functions.rst
```diff
@@ -79,6 +79,8 @@ In our case the alternative version of calling ``parse_date`` would look like th
         set_field("transaction_year", new_date.year);
     end
 
+More examples of the usage of various functions in pipeline rules can be found in the `graylog2-server repo <https://github.com/Graylog2/graylog2-server/tree/master/graylog2-server/src/test/resources/org/graylog/plugins/pipelineprocessor/functions>`_.
```
jalogisch: pointing to the master looks first like a good idea - but it will confuse if the master is way ahead than the current stable release. 

Like it is currently with the master holding 3.0 but stable is 2.4 - just use the link to the current release and this is fine.

jalogisch (changes requested):
Just replace the link to the matching branch to not confuse people.

pages/pipelines/functions.rst
megan201296: @jalogisch maybe it's my inexperience with git and stuff, but I switched to the branch 2.4 and that folder isn't there. You or @joschi know where I can get the same folder but on 2.4 specifically?

jalogisch: That was merged after the 2.4 - the 2.4 has the processing pipelines as an extra plugin.

What was: https://github.com/Graylog2/graylog-plugin-pipeline-processor

So the link would be: https://github.com/Graylog2/graylog-plugin-pipeline-processor/tree/master/plugin/src/test/resources/org/graylog/plugins/pipelineprocessor/functions

megan201296: Ah, thanks!

megan201296:
Made the requested link change @jalogisch

jalogisch (approved):

PR #489, opened 2018-10-10
Update regex example

petersmythe:
It's not obvious from the previous example whether the regex must include the field name or not.  Hopefully this makes it clearer for future users of this documentation.
Ref: https://community.graylog.org/t/searching-via-regular-expression-possible/6008


jalogisch (approved):
LGTM

PR #494, opened 2018-10-25
added updated passgen

jalogisch:
To avoid having the generated password in the bash history the command is updated in the documentation


mpfz0r (changes requested):
echo is missing

mpfz0r (approved):

PR #496, opened 2018-10-26
updated docker versions

jalogisch:
after I noticed that ppl just copy&paste I have adjusted the docker versions


danotorrey (approved):
LGTM 👍

PR #503, opened 2018-11-15
Fix debug log FAQ

danotorrey:
The `curl` command to enable debug logging for a plugin was incorrect (was missing a few arguments). I have fixed and tested it.

screen shot 2018-11-15 at 4 35 53 pm


jalogisch (approved):
LGTM

PR #505, opened 2018-11-30
Add built-in Data Adaptors documentation (mostly for DNS Lookup)

danotorrey:
Add a section in the docs that discusses build-in data adaptors, so that we can have some place in the docs that discusses the DNS Lookup adaptor (@lennartkoopmann will reference from blog post).

image

### DIFF
--- a/pages/lookuptables.rst
+++ b/pages/lookuptables.rst
@@ -176,18 +176,18 @@ of the lookup result, the second returns the **single value**.
 
 .. image:: /images/lookuptables/usage-pipeline-rule.png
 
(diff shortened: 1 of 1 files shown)


bernd (changes requested):
@danotorrey Please use "Adapt**er**" instead of "Adapt**or**"" for consistency with the rest of the documentation. :wink: Thanks!

danotorrey:
@bernd Good eye! Thanks for catching this. Will fix.

jalogisch (approved):
LGTM

PR #510, opened 2018-12-02
Update web_interface.rst

eightnoneone:
include load balancer as means to expose


jalogisch (approved):
LGTM 👍 thank you for this addition.

PR #511, opened 2018-12-04
Update OpenStack docs to include networking setup

cheetahfox:
This change adds a Network section to the OpenStack installation guide. It is very similar to the same section in the AWS guide as a similar issue effects OpenStack instances since most deployments have a VM located in a private Neutron network (similar to an Amazon VPC). 

I ran into this today while trying to use the OpenStack image and when I looked around I saw people talking about this issue going back to May 2017. So I figured putting this section here will save someone else time.


jalogisch (approved):
LGTM 👍 Thank you for your contribution

PR #514, opened 2018-12-07
Adjustments to Integrations documentation

danotorrey:
A few adjustments/enhancement to Enterprise Integrations documentation:

- [*] Move Integrations section to Enterprise > Integrations
- [*] Add setup instructions (for new `graylog-plugin-enterprise-integrations`)
- [*] Improve organization

PR #521, opened 2018-12-14
Fix header underline that is too long

danotorrey:


danotorrey:
Just one cleanup item from #514

PR #525, opened 2018-12-23
Update ubuntu.rst

JimmyKer:
use sed for update /etc/elasticsearch/elasticsearch.yml with cluster.name: graylog


jalogisch:
Thank you for this beautiful idea.

But the documentation should explain what you need to do and not give only copy&paste commands. At least you should try to understand what you are doing.

In addition, when such addition is given, it must have been to all step-by-step guides and not only one.

Thank you again for this, but I'll close this PR and wait for a new one that does not simply replace the setting with the command but give that command as an option to make the changes - and that includes all step-by-step guides.

PR #540, opened 2019-01-24
Add script alert notification documentation

cobaltclaudia:
Documentation for Script Alert Notification.

PR #545, opened 2019-02-05
Fix enterprise tarball download link

bernd:
The download link has changed for rc.1.

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -360,7 +360,7 @@ epub_exclude_files = ['search.html']
 #epub_use_index = True
 
 extlinks = {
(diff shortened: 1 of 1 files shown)


jalogisch:
according to https://github.com/Graylog2/documentation/pull/540 this isn't needed because the download link for tar-balls is on the website only!

bernd:
> according to #540 this isn't needed because the download link for tar-balls is on the website only!

Okay, that means we should remove the link?

bernd:
Ah I see, it got removed already. Closing this.

Thanks for the heads up! :+1:

PR #546, opened 2019-02-05
Add kubernetes master auto selection notes

pastukhov:


jalogisch (changes requested):
could you please use proper [RST Markup](http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) and not MarkDown. 

But thank you for adding this!

pastukhov:
Fixed, i hope

jalogisch:
I have reworded your submission and submitted https://github.com/Graylog2/documentation/pull/548 with that changes. 

This way it was easier to adjust the wording and position. This was just to avoid that we need to play ping-pong with this. 

That is the reason I'll close this PR now. Thank you again for your contribution that will help other users!

PR #547, opened 2019-02-05
Fix script alert config option

danotorrey:
Change server config option from `integrations_web_interface_url` to `integrations_web_interface_uri` so that it matches project standards. Most URLs in server.conf use `_uri` (with an i). The software change for this was implemented in https://github.com/Graylog2/graylog-plugin-enterprise-integrations/pull/27


cobaltclaudia:
LGTM

PR #549, opened 2019-02-06
Increase default content width by 200px

mpfz0r:
The default layout is quite narrow and makes it tedious
to fit our content in tables.
Bump the width from 800px to 1000px.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
-_static
 _build
 
 .DS_Store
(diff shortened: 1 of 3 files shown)


jalogisch (approved):
LGTM 🥇

jalogisch:
Do we want to make that to other branches (2.5) too?

mpfz0r:
> Do we want to make that to other branches (2.5) too?

I guess we could for consistency, yes

PR #553, opened 2019-02-07
Content pack fixup

kmerz:
The nature of content packs has changed dramatically. This change needs to be reflected in the documentaion.


kmerz:
A decision about the Heroku part of the documentation is still needed to finish this PR:
 http://docs.graylog.org/en/2.5/pages/sending_data.html#configuring-graylog-to-receive-heroku-log-messages

jalogisch (approved):
LGTM

we remove the heroku part as long as we do not have something that is working. We can add that later again if we have such a content pack.

PR #558, opened 2019-02-12
Increase default content width by 200px (2nd try)

mpfz0r:
The default layout is quite narrow and makes it tedious
to fit our content in tables.
Bump the width from 800px to 1000px.

(2nd try with different css include)


mpfz0r:
@bernd could we give this another try?

PR #566, opened 2019-02-13
Update API token creation to new UI

mariussturm:
Show the new token UI instead of curl commands for token management.


kmerz (approved):
:taco: :+1:

PR #573, opened 2019-02-15
changed web_listen_uri to http_bind_address, v3 change

digital-grease:
seems like the env var GRAYLOG_WEB_LISTEN_URI has been changed to GRAYLOG_HTTP_BIND_ADDRESS in v3. figured i'd update the doc.


dennisoelkers:
@digital-grease: thanks for the contribution and spotting the stale documentation. As you can see from the [Upgrade Notes](
https://github.com/Graylog2/graylog2-server/blob/master/UPGRADING.rst#simplified-http-interface-configuration), the new setting for `web_endpoint_uri` is `http_external_uri`, so that one should be used instead.

PR #585, opened 2019-02-21
update Prerequisites pkgs

cremesk:
gnupg, gnupg2 and gnupg1 do not seem to be installed, but one of them is required for this operation


jalogisch:
thank you @cremesk while it is not required by Graylog it does not harm to install the package.

But the ubuntu minimal installation contains already all needed packages. I did not find the time yet to verify that. So I'll not merge this as long as I did not tested it.

jalogisch:
I have checked with a Ubuntu LTS 18.04 LTS Server minimal Server installation and `gpg` is already installed. That is why I'll not merge this PR.

If you feel this isn't right, please response and re-open the PR.

PR #586, opened 2019-02-25
Mention new configuration assignment

mpfz0r:
This change has been surprising to a few users.
We should at least document it.


jalogisch (approved):
LGTM

PR #590, opened 2019-03-01
Document breaking change in grouping in ES6

edmundoa:
Document Elasticsearch 6 change in grouping keyword searches, where the `split_on_whitespace` has been removed.

Refs https://github.com/Graylog2/graylog2-server/issues/5694

This also needs to be included in the 2.5 documentation.


jalogisch (approved):
LGTM

thank you for covering this!

PR #591, opened 2019-03-04
updated docker commands

jalogisch:
updated Docker commands

After user feedback all commands are tested and updated. Now additional warning is shown that the commands are only for local usage.


pages/installation/docker.rst
```diff
@@ -360,6 +361,8 @@ Troubleshooting
   In this case please `pick another driver <https://docs.docker.com/engine/userguide/storagedriver/selectadriver>`__ like ``aufs`` or ``overlay``.
 
 
+
+
```
dennisoelkers: Are those two empty lines on purpose?

dennisoelkers (approved):

PR #593, opened 2019-03-06
fix Index mapping curl to include necessary Content-Type Header for P…

shakalandy:
- fix Index mapping curl to include necessary Content-Type Header for PUT requests.
- include link to ES documentation on index mapping

### DIFF
--- a/pages/configuration/elasticsearch.rst
+++ b/pages/configuration/elasticsearch.rst
@@ -74,7 +74,7 @@ Automatic node discovery
 
 .. caution:: Automatic node discovery does not work when using the `Amazon Elasticsearch Service <https://aws.amazon.com/elasticsearch-service/>`_ because Amazon blocks certain Elasticsearch API endpoints.
 
-Graylog uses automatic node discovery to gather a list of all available Elasticsearch nodes in the cluster at runtime and distribute requests among them to potentially increase performance and availability. To enable this feature, you need to set the ``elasticsearch_discovery_enabled`` to ``true``. Optionally, you can define the a filter allowing to selectively include/exclude discovered nodes (details how to specify node filters are found in the `Elasticsearch documentation <https://www.elastic.co/guide/en/elasticsearch/reference/5.4/cluster.html#cluster-nodes>`_) using the ``elasticsearch_discovery_filter`` setting, or tuning the frequency of the node discovery using the ``elasticsearch_discovery_frequency`` configuration option.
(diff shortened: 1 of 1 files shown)


pages/configuration/elasticsearch.rst
```diff
@@ -478,6 +478,7 @@ After you've removed the index template, new indices will only have the original
     }
   }
 
+Additional information on Elasticsearch Index Templates can be found in the official `Elasticseach Documentation <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html>`:
```
jalogisch: would you please write the link following:

```
`Elasticseach Documentation <https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-templates.html>`_
```

In addition we refer linking to a specific version and not current as Elastic tend to change how the product work between versions and we mostly refer to the latest 'official' supported version in Graylog. What is 6.6 at the time of writing.

pages/configuration/elasticsearch.rst
```diff
@@ -338,7 +338,7 @@ Save the following index template for the custom index mapping into a file named
 
 Finally, load the index mapping into Elasticsearch with the following command::
 
-  $ curl -X PUT -d @'graylog-custom-mapping.json' 'http://localhost:9200/_template/graylog-custom-mapping?pretty'
+  $ curl -X PUT -d @'graylog-custom-mapping.json' -H 'Content-Type: application/json' 'http://localhost:9200/_template/graylog-custom-mapping?pretty'
```
jalogisch, line 341: thank you for this catch. 

Internally we moved to using [httpie](https://httpie.org/) that is why such missing bits are unseen.

jalogisch (changes requested):
LGTM just one small change is needed.

jalogisch:
@shakalandy in addition you would need to sign the CLA that we can merge your contribution to the documentation.

pages/configuration/elasticsearch.rst
shakalandy: Updated the link ( and some more to 6.6) hope that is ok.

shakalandy:
I signed the CLA already, seems that my initial push was done with my local git user, and the CLA is not updated after changing to the right "GitHub" one - sorry, i'm a n00b :)

jalogisch:
thank you @shakalandy it might sound childish, but did you mind changing the author of the commit so we can merge?

This might help with that: https://www.git-tower.com/learn/git/faq/change-author-name-email

We can only merge if the CLA check is green, what indicates that each author has signed the CLA, but that is connected to your github handle.

jalogisch:
on request this is closed.

PR #594, opened 2019-03-07
fix example for fuction drop_message

stockf:
The regex function expects the pattern first, then the message

### DIFF
--- a/pages/pipelines/functions.rst
+++ b/pages/pipelines/functions.rst
@@ -814,7 +814,7 @@ Example::
         rule "drop messages over 16383 characters"
         when
             has_field("message") AND
-            regex(to_string($message.message), "^.{16383,}$").matches == true
+            regex(pattern: "^.{16383,}$", value: to_string($message.message)).matches == true
         then
             drop_message();
             // added debug message to be notified about the dropped message


pages/pipelines/functions.rst
```diff
@@ -814,7 +814,7 @@ Example::
         rule "drop messages over 16383 characters"
         when
             has_field("message") AND
-            regex(to_string($message.message), "^.{16383,}$").matches == true
+            regex("^.{16383,}$", to_string($message.message)).matches == true
```
jalogisch: I would prefer if you would include:

```
regex(pattern: "^.{16383,}$",value: to_string($message.message)).matches == true
```

this way it is clear and the order does not matter.

jalogisch (changes requested):

pages/pipelines/functions.rst
stockf: I agree, this is cleaner. I'll make the changes.

stockf:
fixed as required in review

jalogisch (approved):

jalogisch:
thank you @stockf for your contribution.

PR #605, opened 2019-04-02
Correcting word

robsonalves:
in the prhase: "whilst appending", I think the right word is while.


cobaltclaudia (approved):
LGTM. Thank you for your contribution to Graylog.

PR #617, opened 2019-05-05
Update supported Elasticsearch versions

hulkk:
Elasticsearch 7 is not supported.


jalogisch:
thank you for your hint @hulkk we added the information in a little different way to the documentation in https://github.com/Graylog2/documentation/pull/619 that is why I'll close this PR without merging. But thank you again to point to this issue and submitting a fix!
Such is really very welcome!

PR #621, opened 2019-06-04
Refer to correct integrations plugins package

kroepke:
It wrongly referred to the enterprise version of the integrations, but Palo Alto input is in the open source one.


danotorrey (approved):
Good catch!

PR #623, opened 2019-06-18
API Screenshot Updates

r00tnix:
Updated the screenshots on API's to have better resolution


jalogisch (approved):
after removed `scale` the images are working better

PR #624, opened 2019-06-23
revised mongodb install instructions

theDogOfPavlov:
revised mongodb install to use a keyserver approach which works seamlessly through enterprise HTTP proxies etc


jalogisch:
@theDogOfPavlov first let me thank you for this contribution. 

We would like to keep the current version at it is, but make your change as info block available under the current command section. 

Do you mind changing that yourself or should someone of us take care of that?

PR #629, opened 2019-06-27
Update sec_adcs_certificates.rst

Kappty:
INF example has incorrect syntax, and will result in an error.


jalogisch (approved):
LGTM thanks

PR #640, opened 2019-07-15
Add THROTTLED to Documentation

Schuermi7:
A while ago, another LoadBalancer status was introduced but not documented.
So i wrote up the changes, of which I am aware of.


jalogisch (approved):
LGTM

Issue #643, opened 2019-07-17
Malformed formatting in backup.rst


mehrenreich:
## Context

There is a missing space after \`\`server.conf\`\` in line 16.

Issue #642, opened 2019-07-16
Faulty `/var/spool/collector-sidecar/nxlog` reference in sidecar guide


oscarlofwenhamn:
## Context

In the [Graylog Sidecar docs](https://docs.graylog.org/en/3.0/pages/sidecar.html) (specifically the section regarding [NXlog on Ubuntu | CentOS](https://docs.graylog.org/en/3.0/pages/sidecar.html#nxlog-on-centos) collectors) there is a command that I suspect is deprecated:
```sh
$ sudo chown -R nxlog.nxlog /var/spool/collector-sidecar/nxlog
```
The directory in question is no longer present, and I am thinking the corresponding directory may simply be the `/var/spool/nxlog` directory (which might be "owned" by nxlog.nxlog already).

## Expected Content

If the command truly is deprecated, the instructions should be updated with the correct line.

I'd be happy to sort out a PR for this, but I need it confirmed that it is correct first.

Issue #610, opened 2019-04-08
Document Linux kernel requirements for SO_REUSEPORT


bernd:
## Context

Our UDP inputs are using the `SO_REUSEPORT` socket option to allow UDP processing with multiple threads. This socket option has been introduced in the Linux kernel version 3.9.

That means Graylog requires at least Linux kernel 3.9 to run UDP inputs.

Related issue comment: https://github.com/Graylog2/graylog2-server/issues/5825#issuecomment-480822012

## Expected Content

Our documentation should mention the minimum Linux kernel version.

## Your Environment

* Graylog Version: 3.0.1

PR #645, opened 2019-07-17
Take care of issues #610(partly), #642, #643

Schuermi7:
There where 3 issues, which needed a little effort to clean up in the documentation.
The NXLog issue was tested sucessfully with /var/spool/nxlog.  
I suggest, keeping the ``chown`` section, assuring that the file is owned by nxlog but changing the path to the correct version now.
One minor typo was fixed.

### DIFF
--- a/pages/gelf.rst
+++ b/pages/gelf.rst
@@ -50,6 +50,8 @@ Prepend the following structure to your GELF message to make it chunked:
 
 All chunks **MUST** arrive within 5 seconds or the server will discard all already arrived and still arriving chunks. A message **MUST NOT** consist of more than 128 chunks.
 
+Please note, that the UDP-Inputs of Graylog use the ``SO_REUSEPORT`` socket option, which was introduced in Linux kernel version 3.9.  
+So be aware, that UDP inputs will **NOT** work on Linux kernel versions prior to 3.9.
 
 Compression
 -----------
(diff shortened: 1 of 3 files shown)


jalogisch:
May I ask why you closed this? I am on the way review that and does you want to apply a bigger/other PR with this changes?

Schuermi7:
I wanted to rebase on the newest branch and deleted the PR by accident. I am going to restore it.

pages/gelf.rst
```diff
@@ -50,6 +50,8 @@ Prepend the following structure to your GELF message to make it chunked:
 
 All chunks **MUST** arrive within 5 seconds or the server will discard all already arrived and still arriving chunks. A message **MUST NOT** consist of more than 128 chunks.
 
+Please note, that the UDP-Inputs of Graylog use the ``SO_REUSEPORT`` socket option, which was introduced in Linux kernel version 3.9.
```
jalogisch: I would personal made a NOTICE block out of this. Just to make it more visible.

jalogisch (commented):
Thank you for this work on the documentation. I have just a small note - what do you think about that?

Schuermi7:
I added an attention block because I think, that this is a pretty good to know in commercial environments and it is not only worth noticing but worth to get your attention. It was 2 Commits, because at first i didn't understand, how your Boxing-System works.

jalogisch (approved):
LGTM thanks works perfect!

PR #664, opened 2019-07-25
Corrected list of characters requiring escaping

george-viaud:
The list included logical operator strings as opposed to their component characters (eg `||` `&&` vs `|` `&`).


jalogisch (approved):
LGTM

PR #680, opened 2019-08-13
change log 3.1 API path corrections

cawfehman:
modified the changelog 3.1description to eliminate the double slash and clarify the API paths in the new version.


bernd (approved):
Thank you!

PR #703, opened 2019-08-30
Update elasticsearch.rst

### LINKED ISSUE #698: ElasticSearch page hints that ElasticSearch7 is supported
Graylog does not (yet) support ElasticSearch 7. However the page "[Elasticsearch](https://docs.graylog.org/en/3.1/pages/configuration/elasticsearch.html#configuring-es)" does not mention this and even claims that ES7 is supported.
## Context
We installed an incompatible ES7, because the page _Configuring Graylog » Elasticsearch_ claimed

> so it does not have a hard requirement for the Elasticsearch version anymore
> We can safely assume that any version starting from 2.x is working.

and

> Graylog works fine with the Amazon Elasticsearch Service using Elasticsearch 5.3.x or later.

## Expected Content

A caution box with the remark that matches the [Installation/System](https://docs.graylog.org/en/3.1/pages/installation.html#system-requirements) requirements block:

> Graylog 3 does not work with Elasticsearch 7.x!

## Your Environment

* Graylog Version: 3.1 and 3.0
* Elasticsearch Version: 7

jalogisch:
fixes https://github.com/Graylog2/documentation/issues/698


ChristopherKB (commented):
Do we want to leave the previous statement unchanged? It's not really true that it works with ES 5.3 or later. Why not list specific versions (5.3.x through 6.8.x)?

jalogisch:
The statement before is for AWS Elasticsearch - this is generally spoken. 

I do not see a reason for the first one to be changed because I can't verify that for sure.

PR #708, opened 2019-09-06
Fix `cacerts` address in https.rst

Behzadb:
Change `${JAVA_HOME}/jre/lib/security/cacerts1` to `${JAVA_HOME}/lib/security/cacerts`.
`jre` already exist in `java_home` variable.


asachs01:
@Behzadb it's been a while since this request was opened, and I'm not quite sure that this is right. I've checked on a couple of vanilla instances, and in all cases, setting `JAVA_HOME`  doesn't include `jre` as part of the path. That said, if you can show us a system where `jre` is included in the `JAVA_HOME` path, I think we'd be happy to reopen this an accept the PR.

PR #721, opened 2019-09-26
HTTP Webhook Events

jasonkeller:
The structure of these changed dramatically from 3.0 > 3.1, and will break prior parsing code (as it did to me).  Update the documentation to reflect the new structure.


jalogisch (approved):
👍 LGTM Thank you!

PR #722, opened 2019-09-27
ifupdown has been replaced by netplan(5) in ova

cbachert:
Editing "/etc/network/interfaces" on graylog-3.1.2-1.ova gives "ifupdown has been replaced by netplan(5) on this system.  See /etc/netplan for current configuration"

Virtual Machine Appliance documentation should reflect this.


jalogisch (approved):
thank you for this catch!

PR #727, opened 2019-10-09
Removing outdated scripts, using archetype, adding module manual…

dennisoelkers:
This change is removing references to outdated scripts which are confusing new users. The generated plugin is added manually to the POM for now (until we have support in `graylog-project` for that).

### DIFF
--- a/pages/plugins.rst
+++ b/pages/plugins.rst
@@ -75,10 +75,7 @@ Use it like this::
   $ git clone git@github.com:Graylog2/graylog-project.git
 
 
-This will create a checkout of the meta project in your current work dir. Now change to the ``graylog-project`` directory and execute the step which to download the necessary base modules::
-
-  $ graylog-project bootstrap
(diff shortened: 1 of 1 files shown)


pages/plugins.rst
```diff
@@ -77,12 +77,12 @@ Use it like this::
 
 This will create a checkout of the meta project in your current work dir. Now change to the ``graylog-project`` directory and execute the step which to download the necessary base modules::
 
-  $ scripts/bootstrap
+  $ graylog-project bootstrap
```
linuspahl: In the graylog-projects readme, the bootstrap command is followed by the repo url `github://Graylog2/graylog-project.git`. Should we add this here as well?

jalogisch (approved):
from a logical and technical point all right - waiting on resolve/approve of @linuspahl

linuspahl (approved):

PR #737, opened 2019-11-11
Correct example for route_to_stream function

### LINKED ISSUE #6259: Is route_to_stream documented incorrectly?
The documentation page describes arguments this way:
https://docs.graylog.org/en/latest/pages/pipelines/functions.html#route-to-stream 

`route_to_stream(id: string | name: string, [message: Message], [remove_from_default: boolean])`

(so there is only one argument for the stream identification)
it also contains examples supporting this way of usage.

But the "Rules quick reference" section in the Rule editor describes the syntax differently:

`route_to_stream([name], [id], [message], [remove_from_default]) : Void`

(there are two arguments now - one for the name and one for the id)

From my experience, only the second variant works.

#### My Environment

* Graylog 3.0.0+db6cf59 on 183a1a5b6ff1 (Oracle Corporation 1.8.0_181 on Linux 4.14.101-91.76.amzn2.x86_64)

thll:
fixes Graylog2/graylog2-server#6259

The example given for the route_to_stream function doesn't work, at least not for routing by ID. The example for routing by name did work, but I changed it to be more obvious and in line with routing by id.

### DIFF
--- a/pages/pipelines/functions.rst
+++ b/pages/pipelines/functions.rst
@@ -798,10 +798,10 @@ If ``remove_from_default`` is ``true``, the message is also removed from the def
 Example::
 
         // Route the current processed message to a stream with ID `512bad1a535b43bd6f3f5e86` (preferred method)
-        route_to_stream("512bad1a535b43bd6f3f5e86");
+        route_to_stream(id: "512bad1a535b43bd6f3f5e86");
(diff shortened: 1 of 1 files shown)


jalogisch (approved):
nice catch! did not see that

PR #745, opened 2019-11-21
Add important note about version

LeeU1911:
Versions of Graylog and `graylog-integrations-plugins` must be the same


jalogisch (approved):
Thank you for this.

PR #766, opened 2020-01-13
Add documentation for enabled_tls_protocols settings

mpfz0r:
and retire jdk specific ciphers and algorithms chapter


jalogisch (changes requested):
we should add some information about the sane defaults we decide to support. Because that is the first we will be asked in the community and in the support. 

So at least at the time of writing the documentation or by linking to the code that people that are willing and able to can take the current status out. 

When we went silent on that we will get questions and that end up in a section of the FAQ or similar.

jalogisch (approved):
thx LGTM

PR #772, opened 2020-01-23
8192 bytes is not applying to UDP

sfuhrm:
8192 bytes is not applying to UDP (see https://tools.ietf.org/html/rfc768), but to the application. Added a clarification for that.


jalogisch:
@sfuhrm do you mind signing the CLA? only with that we are able to merge your contribution.

jalogisch (approved):

PR #776, opened 2020-01-24
Add links to step by step guides for various os packages

cobaltclaudia:


pages/installation/operating_system_packages.rst
```diff
@@ -9,11 +9,7 @@ software on Linux servers was to use operating system packages. Debian has ``DEB
 based on those or come with their own package formats. Online repositories of software packages and corresponding package managers make installing
 and configuring new software a matter of a single command and a few minutes of time.
 
-Graylog offers official ``DEB`` and ``RPM`` package repositories. The packages have been tested on the following operating systems:
-
-* Ubuntu 16.04, 18.04
-* Debian 8, 9
-* RHEL/CentOS 6, 7
+Graylog offers official ``DEB`` and ``RPM`` package repositories for the operating systems listed below in the **Step by Step Guides** section.
```
jalogisch: the installation steps for some OS installations differ - so it has to be different packages for different ubuntu versions, same for debian or centos. 

We do active write in the step-by-step guide for what version this is written, but that does not mean it will work for other versions also. We should be careful how we write for what versions this work, because people count on that.

jalogisch (changes requested):
🥇 thx for this 

we should be careful about the version numbers and be conversative about what the official wording is in the docs.

jalogisch (approved):
💯 👍

PR #779, opened 2020-01-29
Add more upgrade notes

bernd:
This is the content from https://github.com/Graylog2/graylog2-server/blob/3.2/UPGRADING.rst

### DIFF
--- a/pages/upgrade/graylog-3.2.rst
+++ b/pages/upgrade/graylog-3.2.rst
@@ -9,6 +9,66 @@ Upgrading to Graylog 3.2.x
    :backlinks: top
 
 
+.. note:: Graylog 3.2.0 comes with a number of migrations that change parts of your existing data fundamentally. Please make sure to have a recent backup of your MongoDB that you can rollback to, before attempting to upgrade from an earlier version.
+
+Migrating Dashboards
+====================
+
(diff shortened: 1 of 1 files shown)


mpfz0r (dismissed):
LGTM, someone else should double check the Dashboard parts

pages/upgrade/graylog-3.2.rst
```diff
@@ -9,6 +9,64 @@ Upgrading to Graylog 3.2.x
    :backlinks: top
 
 
+.. note:: Graylog 3.2.0 comes with a number of migrations that change parts of your existing data fundamentally. Please make sure to have a recent backup of your MongoDB that you can rollback to, before attempting to upgrade from an earlier version.
```
jalogisch: I would make a warn and not only a note

jalogisch (dismissed):
👍 

Only a single item I would change. But that is not fundamental at all.

bernd:
> LGTM, someone else should double check the Dashboard parts

The dashboards parts have been written by the search team, so I guess they are fine. :smiley:

pages/upgrade/graylog-3.2.rst
bernd: I updated the PR to make this a warning.

mpfz0r (approved):

PR #785, opened 2020-02-03
Update configuration/server.conf for 3.2

bernd:
This adds new config file options and adjusts some changed defaults.


jalogisch (approved):
👍

PR #788, opened 2020-02-04
fix integration plugins version mentioned

jalogisch:
The integrations setup page has a wrong version number mentioned and people follow that version mentioned:

https://community.graylog.org/t/log-forwarding-between-2-graylog-servers/13749

So I have changed this and added the note about the version need to be the same as Graylog server.


cobaltclaudia (approved):
LGTM 👍

PR #789, opened 2020-02-04
update centos install page with yum instead of apt-get

z3r0bytes:
changed apt-get to yum


jalogisch (approved):
👍 

good catch - during the update of the docs this was slipped in ... thank you for this.

PR #796, opened 2020-02-07
Update saved search image paths

linuspahl:
It looks like replacing an image has no effect and you need to adjust the image name as well. This is maybe related to the cdn cache.

This PR updates the names for two images on the saved search page.


jalogisch (approved):
LGTM 👍

PR #799, opened 2020-02-08
Fix typo in ubuntu install

cobaltclaudia:


jalogisch (approved):

jalogisch:
I'm not sure about this - we should return to the chained command.

PR #801, opened 2020-02-09
Add info about displaying syntax errors and updating images to readme

linuspahl:
While adding changes for the documentation I've learned two things
- the CI errors will not be displayed locally by default
- updating an image only affects the production version immediately when the image name changes

This PR adds this information to the readme.


jalogisch (approved):
👍

PR #803, opened 2020-02-10
clarify on index template

jalogisch:
This is to avoid confusion: https://community.graylog.org/t/upgrading-to-graylog-3-2-psa/13836


linuspahl (approved):
LGTM!

PR #820, opened 2020-02-24
Http json input

lingpri:
added an example for input, that would matches the curl command and the message.


lingpri:
@jalogisch @ceruleancee -  Please see the local validation of the document and let me know if you have any questions..
doc_jsoninput_valiadate

jalogisch (approved):

PR #822, opened 2020-02-27
Lookup table mongodb adapter

kmerz:
The lookup table adapter for mongodb did not yet got mentioned in the documentation.

To have a reference for users and customers I added a section.


jalogisch (approved):
<3 👍 

Love this function!

PR #823, opened 2020-03-03
made clear functions for mongodb lut

jalogisch:
It wasn't clear that lookup table manipulation is only possible (at the moment) for lookup tables in mongodb. 

This PR should make this more clear.


kmerz (approved):
Thanks for clearing that! :+1:

PR #829, opened 2020-03-09
Add page about search parameters

linuspahl:
This PR adds a page about parameters, which is part of the search section.


pages/searching/parameters.rst
```diff
@@ -0,0 +1,38 @@
+Parameters
+----------
+
```
jalogisch, line 5: we need to mention that this is an enterprise feature.

pages/searching/parameters.rst
```diff
@@ -0,0 +1,38 @@
+Parameters
```
jalogisch, line 3: add before:

```
.. _search_parameters:

```

That we have the ability to use references to this page by that.

jalogisch (changes requested):
like it!

only two minor comments.

jalogisch (approved):
👍

PR #843, opened 2020-03-31
Rename widget field action "Aggregate" to "Show top values"

linuspahl:
This PR updates the field action titles and the related screenshot.


jalogisch (changes requested):
some of the text in between does use the phrase "aggregation" too - maybe you can quick check this too?

linuspahl:
@jan The text I can find relates to the widget type "Aggregation", can you give me an example where "Show top values" makes more sense than "Aggregation"?

We only renamed the field action, not the widget type.

jalogisch (approved):
thanks @linuspahl 

might look into any inconsistence sentences later.

Issue #870, opened 2020-05-13
Event notification documentation is outdated


mpfz0r:
The documentation does not match the new alerting system we've introduced with 3.1.

Specifically the email templates has changed completely and is wrong:
 https://docs.graylog.org/en/3.2/pages/alerts.html#email-alert-notification

The http notification also needs an update to reflect the changed format.

jalogisch:
https://github.com/Graylog2/documentation/pull/876 has closed this

PR #871, opened 2020-05-13
Remove old alert notification documentation

mpfz0r:
The old docs were wrong and confusing.
It's better to have no documentation than wrong documentation.

Refs #870


jalogisch (approved):
better having this than wrong

jalogisch:
do we backport this to 3.2 docs too?

mpfz0r:
> do we backport this to 3.2 docs too?

I guess that's reasonable, I'll do it

PR #872, opened 2020-05-13
Fix the docker run command

mpfz0r:


jalogisch (approved):
👍 thank you for fixing

PR #886, opened 2020-06-09
Update setup.rst

DeyV:
https://docs.graylog.org/en/3.3/pages/integrations/setup.html  - 3.3 version of plugins


jalogisch (approved):
🌮 

thank you for this catch.

PR #890, opened 2020-06-10
Squadcast http notification

raghuchinnannan:
Added a section for HTTP Notification Alerts which are officially supported by other platforms and added Squadcast as the first entry.

This list can keep growing as more vendors starts supporting it and the documentation can be updated.

### DIFF
--- a/pages/alerts.rst
+++ b/pages/alerts.rst
@@ -243,9 +243,9 @@ HTTP alert notification - Supported platforms
 ---------------------------------------------
 
 Squadcast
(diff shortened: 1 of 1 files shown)


jalogisch:
he @raghuchinnannan 

thank you for this contribution. But we can not accept this as it would make the impression that Graylog does not work with any other service. Or we would need to add more information about other services. 
As we do not want to give recommendations here and do not have the manpower to make all kind of tests with different vendors we will not include this. 

thank you again.

PR #905, opened 2020-07-09
Update conf.py

grownuphacker:
Updated version number.  The documentation for 3.3 says "3.2" across the title bar and far more importantly on Google Search results as its the page title on all the 3.3 docs.  I am simply guessing at the location of where the '3.2' header is coming from.  Truthfully, I have no idea if this will break anything or not!


grownuphacker:
Browser View: 

image

Google Results look pretty odd: 
image

jalogisch (approved):

PR #911, opened 2020-07-21
fixed wording

### LINKED ISSUE #865: docker-compose example suggest mounting volume at wrong location
[This example docker-compose file](https://github.com/Graylog2/documentation/blob/3.2/pages/installation/docker.rst) suggests to mount a volume called `graylog_journal` at `/usr/share/graylog/data/journal` inside the Docker container.

```
    volumes:
      - graylog_journal:/usr/share/graylog/data/journal
```

However, the actual mounted location according to [this Dockerfile](https://github.com/Graylog2/graylog-docker/blob/3.2/Dockerfile#L127) is 

`/usr/share/graylog/data`

This is causing Docker to make new volumes at `/usr/share/graylog/data` every time the container is restarted.
Because of this, all the contents of `/usr/share/graylog/data` are **lost on reboot** (except the `journal` directory, which is the mounted volume).

Below is the contents of `/usr/share/graylog/data`, for reference.
```
graylog@7fbd7f50a312:~$ ls -la /usr/share/graylog/data/
total 36
drwxr-x--- 9 graylog graylog 4096 May  2 02:46 .
drwxr-xr-x 1 graylog graylog 4096 Mar 19 19:18 ..
drwxr-x--- 2 graylog graylog 4096 May  2 02:46 config
drwxr-xr-x 2 graylog graylog 4096 May  2 02:46 contentpacks
drwxr-x--- 2 graylog graylog 4096 Mar 19 19:18 data
drwxr-x--- 3 graylog graylog 4096 May  2 02:56 journal
drwxr-xr-x 2 graylog graylog 4096 May  2 02:46 libnative
drwxr-x--- 2 graylog graylog 4096 Mar 19 19:18 log
drwxr-x--- 2 graylog graylog 4096 Mar 19 19:18 plugin
```

## Context

I noticed my node_id was changing after I restarted the Docker container. This caused my inputs to not start (because they were built with old node_id).

https://community.graylog.org/t/an-input-has-failed-to-start-but-all-my-inputs-are-already-running/15226

## Expected Content

I believe the docker-compose example should be this. I set this up and now my inputs are starting and my node_id is the same every time the container starts.

```
    volumes:
      - graylog_data:/usr/share/graylog/data
```

Can you confirm this? For this to affect you, you would have to **not** have the node-id file mounted ([like this person does](https://github.com/jalogisch/d-gray-lab/blob/master/docker-compose.yml#L304)).

## Your Environment

* Graylog Version: 3.2
* Elasticsearch Version: 6.8.8
* MongoDB Version: 4.0.18
* Operating System: Docker (Ubuntu 18.04 host)
* Browser version: Firefox 75

jalogisch:
After thinking about https://github.com/Graylog2/documentation/issues/865 and the knowledge that we not only have the journal that you want to be persistent.

FIX: https://github.com/Graylog2/documentation/issues/865


kmerz (approved):
LGTM :+1:

PR #912, opened 2020-07-21
change to copy&paste action

jalogisch:
I just realised that we try to have as many copy&paste options as possible. 

With this change we have removed the reference to vim to edit a file and just added the option to copy&paste from the docs. 

Hopefully this will lower the amount of people forgetting the setting.


kmerz (approved):
:+1: LGTM

PR #920, opened 2020-07-24
Add documentation for Enterprise Output Framework

waab76:
This is the initial version of the documentation for the Enterprise Output Framework developed for release in 3.3.3.


pages/integrations/output_framework.rst
```diff
@@ -0,0 +1,142 @@
...
+----------------------------
+
+The Enterprise Output Framework provides a number of new Outputs for the various network 
+transport types. All of these Outputs first write messages to an on-disk journal in the 
+Graylog cluster.  Messages stay in the on-disk journal until the Output is able to 
+successfully send the data to the external receiver.
+
+Once the messages have been written to the journal, they are optionally run through a 
+processing pipeline so unwanted messages can be filtered out and the remaining messages can 
+be modified or enriched with additional data.
```
r00tnix: Maybe re-word this a bit:   (Just so the filter is not the first thing :))

processing pipeline to modify or enrich logs with additional data, transform the message contents or filter out any logs before sending.

pages/integrations/output_framework.rst
```diff
@@ -0,0 +1,142 @@
...
+  disk journal
+- **Journal Buffer Size**: The size of the memory buffer for messages waiting to be written 
+  to the journal. This must be a power of two.
+- **Journal Buffer Encoders**: The number of concurrent encoders for messages being written
+  to the journal.
+- **Output Processing Pipeline**: The pipeline which will process all messages sent to the
+  Output
+- **Outbound Payload Format**: The format that will be used for outgoing message payloads
+
+
```
r00tnix: Do we have any notes, on where the Journal will be stored at by default?  Just for documentation sake so they know where to look at.

r00tnix (changes requested):
Added a couple of notes below.

pages/integrations/output_framework.rst
waab76: Will do

pages/integrations/output_framework.rst
waab76: I believe all on-disk journal data is stored in the same location specified by the `message_journal_dir` value in the `graylog.conf` file.

pages/integrations/output_framework.rst
```diff
@@ -0,0 +1,146 @@
...
+----------------------------
+
+The Enterprise Output Framework provides a number of new Outputs for the various network 
+transport types. All of these Outputs first write messages to an on-disk journal in the 
+Graylog cluster.  Messages stay in the on-disk journal until the Output is able to 
+successfully send the data to the external receiver.
+
+Once the messages have been written to the journal, they are optionally run through a 
+processing pipeline to modify or enrich logs with additional data, transform the message
+contents, or filter out any some logs before sending.
```
jalogisch, line 30: For clarification - the data is written to the journal first, then send to the processing pipeline for modification and send to the selected output. 

Or is the output modified by the processing pipeline, written to the journal, read in and modified to the desired output format and send to the target?

pages/integrations/output_framework.rst
```diff
@@ -0,0 +1,146 @@
...
+rates. Many hardware factors will affect throughput (such as CPU clock speed, number of 
+CPU cores, available memory, and network bandwidth). Several Output Framework configuration 
+options are available to help you tune performance for your throughput requirements and 
+environment.
+
+Common Options
+^^^^^^^^^^^^^^^
+
+.. image:: /images/integrations/output-framework-config-01.png
+.. image:: /images/integrations/output-framework-config-02.png
```
jalogisch: I would cut out those images - and have only the list. The images do not make any sense as the information is the same as the list.

pages/integrations/output_framework.rst
```diff
@@ -0,0 +1,146 @@
...
+- **Output Processing Pipeline**: The pipeline which will process all messages sent to the
+  Output
+- **Outbound Payload Format**: The format that will be used for outgoing message payloads
+
+
+
+TCP Raw/Plaintext Options
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+.. image:: /images/integrations/output-tcp-options.png
```
jalogisch: same as for the other image.

pages/integrations/output_framework.rst
```diff
@@ -0,0 +1,146 @@
...
+  end of the message. Any newline characters within the message will be escaped prior to sending.
+- **Null Character**: A null character will be appended to each message to mark the end of 
+  the message. Any null characters within the message will be escaped prior to sending.
+- **Octet Counting**: The length of the message (in bytes) and a space character for 
+  separation will be prepended to the message.  The contents of the message will not be altered.
+
+TCP Syslog Options
+^^^^^^^^^^^^^^^^^^
+
+.. image:: /images/integrations/output-tcp-options-syslog.png
```
jalogisch: same as with the other images.

pages/integrations/output_framework.rst
```diff
@@ -0,0 +1,146 @@
...
+
+The Facility and Severity fields are defined in `Section 6.2.1 <https://tools.ietf.org/html/rfc5424#section-6.2.1>`_ of the Syslog specification.
+
+- **Syslog Facility**: A numeric value in the range of 0 - 23 (inclusive)
+- **Syslog Severity**: A numeric value in the range of 0 - 7 (inclusive)
```
jalogisch: not all options from the image are described.

jalogisch (changes requested):
I would personal adjust the styling to the same as we have used in:

pages/configuration/server.conf.html

this gives a little more structure. 

In addition I would remove the input images, but add one additional processing image. 

Once the first question is clarified I can provide that image.

pages/integrations/output_framework.rst
```diff
@@ -40,6 +40,10 @@ the external receiver. This allows the Output Framework to keep receiving and re
 queuing messages, even if the external receiver is temporarily unavailable due to
 network issues. The journal has many configuration options which will be described below.
 
+By default, journal data for Framework Outputs will be stored in the same directory used
+for the Input journal.  This directory is controlled by the ``message_journal_dir`` value
```
r00tnix, line 48: Thanks for the update here.  One more question on it, will there be a specific name for this journal? If so, we could note it, so they could keep track of it.

pages/integrations/output_framework.rst
waab76: The flow goes like this:

1. Outgoing message is written to the journal
1. Journal handler executes optional pipeline
1. Journal handler formats message
1. Journal handler sends formatted message
1. Journal handler updates journal to indicate that the message is done being handled

pages/integrations/output_framework.rst
waab76: Ok, I'll cut out the images

pages/integrations/output_framework.rst
waab76: Will do

pages/integrations/output_framework.rst
waab76: Will do

pages/integrations/output_framework.rst
waab76: Other options in the image were described previously

pages/integrations/output_framework.rst
waab76: I think the journal name is randomly chosen but specific to the input.

pages/integrations/output_framework.rst
cobaltclaudia: adding a diagram of the flow of the data might assist in any confusion a user might have, similar to the way docs for the [forwarder](https://docs.graylog.org/en/3.3/pages/integrations/forwarder.html) are.

waab76: Updated some styling and removed images on latest push.  Jan will work on the diagram and do some additional styling work

pages/integrations/output_framework.rst
```diff
@@ -0,0 +1,144 @@
...
+
+Output Configuration
+--------------------
+
+The Enterprise Output Framework is capable of processing messages at very high throughput 
+rates. Many hardware factors will affect throughput (such as CPU clock speed, number of 
+CPU cores, available memory, and network bandwidth). Several Output Framework configuration 
+options are available to help you tune performance for your throughput requirements and 
+environment.
+
```
cobaltclaudia, line 93: adding a list of the three different configuration types (with anchors) would be helpful to the user while navigating the docs

cobaltclaudia (dismissed):
LGTM

jalogisch (dismissed):
first version - GO

I would make the options more verbose but that should not lead to a delayed documentation.

pages/integrations/output_framework.rst
jalogisch: we might want to add this as additional push later on.

jalogisch (dismissed):
👍

PR #947, opened 2020-09-23
Add Cloudtrail to list of throttled inputs

mariussturm:
Just another throttled input. Was searching for it but couldn't find it in the docs.


jalogisch (approved):
👍

PR #949, opened 2020-09-24
Documenting ES node discovery options.

dennisoelkers:
This change is documenting the newly added default user/password options for ES node discovery, as well as the previously existing default scheme.


jalogisch (approved):
👍

PR #954, opened 2020-10-01
added  dejavu font

### LINKED ISSUE #937: update installation packages to include fonts
## Context
we noticed that some installations are missing fonts to generate proper reports.

The solution was to install `dejavu-sans-fonts` and `dejavu-serif-fonts`. This should be added to the enterprise section that the reports needs those packages.

jalogisch:
added the installation of the font package as some minimal installations does not have that package.

FIX: #937

needs to be pushed to 4.0 too


dennisoelkers (approved):
Thanks for the change!

PR #955, opened 2020-10-01
added warn and link 

### LINKED ISSUE #924: first sample docker-compose.yml should persist data
## Context
I set up Graylog in Docker using the first sample docker-compose.yml in the documentation (https://docs.graylog.org/en/3.3/pages/installation/docker.html). It was only later I read the "Persisting data" section, and lost all my Elasticsearch data.

The Graylog and Mongo docker images use volumes, so docker-compose created unnamed volumes and no data is lost if you recreate the container. However the elasticsearch-oss image does not use any volumes, so recreating the volume destroys all your data. 

The slightest tweak to the settings for it in docker-compose causes it to be recreated, so even adding the named volume later causes your data to be deleted!

## Expected Content
Data should not be lost so easily. I suggest just including the named volumes in the first sample docker-compose.yml presented in the documentation. Is there any reason not to?

## Your Environment

* Graylog Version: 3.3
* Operating System: Linux

jalogisch:
decided to add a warning and a link to the persistent data section. 

FIX: #924

needs to be pushed to 4.0 too


asachs01 (approved):
@jalogisch LGTM 👍

PR #960, opened 2020-10-07
Updating documentation for Enterprise Output journal

waab76:
Updating the On-Disk Journal section of the output framework docs to correct how the on-disk location is determined and provide guidance on managing the size/growth of the on-disk journal.


jalogisch (approved):
That Looks Good to me - please do not forget to merge that into 4.0 branch of the docs

PR #969, opened 2020-10-19
Adds supplemental instruction for Docker quick start

### LINKED ISSUE #968: Docker instructions missing critical information needed for success
## Context

In the [Docker quick start examples](https://docs.graylog.org/en/3.3/pages/installation/docker.html#how-to-get-log-data-in), there is some information missing from the instructions. Namely, that in order to get log data into the Docker container, there must be an accompanying Raw/Plaintext input created in the Graylog UI before a user can use the example command `echo 'First log message' | nc localhost 5555` to generate test data inside the UI. 

## Expected Content

Under the "Getting log data in" section, there is this bit:

> You can create different kinds of inputs under System / Inputs, however you can only use ports that have been properly mapped to your docker container, otherwise data will not go through.

In this case, we should be very explicit about what is needed. As this is a quick start, we have to be explicit so that any user going through the scenario is able to get immediate value.

asachs01:
This PR fixes #968 by adding screenshots and clarifying language around what is needed in the Docker quickstart so that users can see immediate value when following the Docker examples.


waab76 (dismissed):
LGTM

asachs01:
@waab76 can you approve again please? I had an uncommitted change that I hadn't pushed up.

waab76 (approved):

PR #971, opened 2020-10-22
Adding ES rolling upgrade notes.

dennisoelkers:
This change is adding notes when and how rolling upgrades of the
Elasticsearch cluster are supported and suggests a procedure to be
followed for major version upgrades of the ES cluster without suffering
from message loss.

### DIFF
--- a/pages/upgrade/rolling_es_upgrade.rst
+++ b/pages/upgrade/rolling_es_upgrade.rst
@@ -12,7 +12,7 @@ Graylog supports rolling upgrades without restarting any Graylog node for Elasti
 
 To avoid message loss in case of a rolling Elasticsearch upgrade which requires the restart of Graylog nodes, please follow this procedure:
 
-1. Temporarily disable ES automatic node discovery if it is used.
+1. Temporarily disable ES :ref:`automatic node discovery <automatic_node_discovery>`_ if it is used.
 2. Mentally split up the Elasticsearch cluster in two groups by designating each Elasticsearch node to group A or B.
 3. Mentally split up the Graylog cluster in two groups by designating each Graylog node to group A or B.
 4. Temporarily configure each Graylog node from group A to use only Elasticsearch nodes from group A and vice versa for group B.
--- a/pages/upgrade/rolling_es_upgrade.rst
+++ b/pages/upgrade/rolling_es_upgrade.rst
@@ -16,12 +16,12 @@ To avoid message loss in case of a rolling Elasticsearch upgrade which requires 
 2. Mentally split up the Elasticsearch cluster in two groups by designating each Elasticsearch node to group A or B.
 3. Mentally split up the Graylog cluster in two groups by designating each Graylog node to group A or B.
 4. Temporarily configure each Graylog node from group A to use only Elasticsearch nodes from group A and vice versa for group B.
-5. Shut down Graylog nodes from group B.
+5. Pause message processing on each Graylog node from group B _or_ shut down Graylog nodes from group B.
 6. Upgrade all Elasticsearch nodes from group B.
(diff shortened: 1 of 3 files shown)


pages/upgrade/rolling_es_upgrade.rst
```diff
@@ -0,0 +1,32 @@
...
+
+This page contains a few notes and a recommended procedure to perform a rolling upgrade for a Elasticsearch cluster utilized by Graylog.
+
+Elasticsearch supports rolling upgrades to avoid downtimes during upgrades. Detailed information about the procedures and limitations are provided `here <https://www.elastic.co/guide/en/elasticsearch/reference/master/rolling-upgrades.html>`__.
+
+Graylog supports rolling upgrades without restarting any Graylog node for Elasticsearch upgrades *between minor versions*. While Elasticsearch supports upgrading from e.g. 6.8 to 7.0, Graylog requires a restart when the major version of the Elasticsearch cluster is changed.
+
+To avoid message loss in case of a rolling Elasticsearch upgrade which requires the restart of Graylog nodes, please follow this procedure:
+
+1. Temporarily disable ES automatic node discovery if it is used.
```
kroepke: I'd link this to `https://docs.graylog.org/en/latest/pages/configuration/elasticsearch.html#automatic-node-discovery` so people know which discovery is meant.

pages/upgrade/rolling_es_upgrade.rst
```diff
@@ -0,0 +1,32 @@
...
+Elasticsearch Rolling Upgrade Notes
+***********************************
+
+This page contains a few notes and a recommended procedure to perform a rolling upgrade for a Elasticsearch cluster utilized by Graylog.
+
+Elasticsearch supports rolling upgrades to avoid downtimes during upgrades. Detailed information about the procedures and limitations are provided `here <https://www.elastic.co/guide/en/elasticsearch/reference/master/rolling-upgrades.html>`__.
+
+Graylog supports rolling upgrades without restarting any Graylog node for Elasticsearch upgrades *between minor versions*. While Elasticsearch supports upgrading from e.g. 6.8 to 7.0, Graylog requires a restart when the major version of the Elasticsearch cluster is changed.
+
+To avoid message loss in case of a rolling Elasticsearch upgrade which requires the restart of Graylog nodes, please follow this procedure:
```
kroepke, line 13: Would this scenario be a little easier if we told people to stop message processing?
I think in that case we could wholesale update the ES cluster in one go, and then restart the Graylog nodes one-by-one to ensure message delivery.

Afterwards, enabling processing again should not lose any messages as well, if at the cost of having a brief interruption in ability to search. Maybe mention it as a second option?

kroepke (changes requested):
In general looks good, save for the link to the other section.

The other comment is more a suggestion than anything else.

pages/upgrade/rolling_es_upgrade.rst
dennisoelkers: Yes, that would work as well. Shutting down the Graylog node just does not need an explanation. But maybe we have a good part in the docs I could link to for stopping message processing?

pages/upgrade/rolling_es_upgrade.rst
dennisoelkers: :+1:

pages/upgrade/rolling_es_upgrade.rst
kroepke: hm, appears we don't actually. Let's keep it the way it is, probably better to have a single option than multiple ones that confuse the matter further.

kroepke (approved):
lgtm thanks!

PR #976, opened 2020-10-30
update Slack plugin

lingpri:
alerts_slack_notification


ChristopherKB (approved):
Looks OK to me. I've never used the old plugin, so I don't have much to offer in the way of feedback until I get to use it.

pages/alerts.rst
```diff
@@ -284,11 +284,57 @@ Client URL
...
+
+Configuration Color
+    Highlight the custom message with this `color <https://api.slack.com/reference/surfaces/formatting#quotes>`__ .
+
+Title
+    A title for this notification.
+
+Description
+    A long description for this slack notification.
+
```
waab76, line 312: These are common to all Notifications and don't need to be included for each one.

pages/alerts.rst
```diff
@@ -284,11 +284,57 @@ Client URL
...
 
 
+.. _alert_notification_Slack Notification:
+
+Slack alert notification
+------------------------
+The Slack alert notification allows you to send notifications using your `Webhook URL <https://api.slack.com/messaging/webhooks>`__ to your slack workspace
+in response to the configured event definitions in your Graylog server.
+
+.. image:: /images/alerts_slack_notification.png
```
waab76, line 294: Screen shot is very wide, which means the text comes out very small on the page.  Could we get a narrower screen shot?

pages/alerts.rst
```diff
@@ -284,11 +284,57 @@ Client URL
...
+
+These are the supported configuration options:
+
+Webhook URL
+    An unique URL to which you send a JSON payload with the message text and some options.
```
waab76, line 299: I think this can just say "The unique URL used to send messages to your Slack instance."  Also, might want to move the URL link here instead of up in the description.

pages/alerts.rst
```diff
@@ -284,11 +284,57 @@ Client URL
 .. image:: /images/alerts_pagerduty_notification.png
 
 
+.. _alert_notification_Slack Notification:
+
+Slack alert notification
+------------------------
+The Slack alert notification allows you to send notifications using your `Webhook URL <https://api.slack.com/messaging/webhooks>`__ to your slack workspace
```
waab76, line 291: I think the URL link would make more sense below in the configuration option for the Webhook URL.  In fact, this might read better as "The Slack alert notification allows you to send notifications to your slack workspace in response to events in your Graylog server."

pages/alerts.rst
```diff
@@ -284,11 +284,57 @@ Client URL
...
+    Highlight the custom message with this `color <https://api.slack.com/reference/surfaces/formatting#quotes>`__ .
+
+Title
+    A title for this notification.
+
+Description
+    A long description for this slack notification.
+
+Custom Message
+    A custom message to be appended below the alert title.
```
waab76, line 314: Might add something here indicating that this is a message template and the "data described above" (several notifications use this phrase to refer to the `Data available to notifications` section of the page) can be used.

pages/alerts.rst
```diff
@@ -284,11 +284,57 @@ Client URL
...
+
+User Name (optional)
+    Addressing a specific `user  <https://api.slack.com/reference/surfaces/formatting#mentioning-users>`__ in the slack workspace.
```
waab76, line 317: If I'm not mistaken, this is the user that the messages in Slack will appear to come from, not a user to be mentioned or tagged in the message as the link indicates.

pages/alerts.rst
```diff
@@ -284,11 +284,57 @@ Client URL
...
+
+Icon URL
+    Image to use as the icon for this message.
+
+Icon Emoji
+    `:turtle:`
```
waab76, line 323: This is not a very good description of this field.

pages/alerts.rst
lingpri: https://github.com/Graylog2/documentation/pull/986

pages/alerts.rst
lingpri: notification_username

the user is mentioned above in the picture. The notification is always within the context of the event definition. Hope it's clear to you.

waab76: On the config page for the Slack plugin, this field is described as "User name of the sender in Slack".  I think the same applies here.

PR #995, opened 2020-11-12
add backlog message items field to Slack Plugin

lingpri:
slack_alert_docs


jalogisch:
he @lingpri looks good but will you fix the build error or shall I do?

pages/alerts.rst
```diff
@@ -312,6 +314,13 @@ Configuration Color
 Custom Message
     The message that will be sent to Slack.  The data described above can be used in this template.
 
+Backlog Message Items (optional)
```
waab76: Name here should match whatever is used across the 3 pages in the other PR.

pages/alerts.rst
```diff
@@ -312,6 +314,13 @@ Configuration Color
...
+    The custom message has a data point named `backlog <http://localhost:8000/pages/alerts.html?highlight=alerts#backlog>`__, which gives
+    the preview of the messages that led to a slack notification.
```
waab76: I don't think we need to describe here what the message backlog is.  That should be discussed elsewhere in the Alert documentation.

pages/alerts.rst
```diff
@@ -312,6 +314,13 @@ Configuration Color
...
+    Backlog Message Items will be used to override the Event definition Notification settings field `Message Backlog`
```
waab76: Again, name should match what is used on the UI (and I think should include the word `Limit` to make the purpose clear).

pages/alerts.rst
```diff
@@ -312,6 +314,13 @@ Configuration Color
...
+    to limit the number of backlog messages sent as part of the slack message.
+    A Slack notification cannot have more than 50 backlog messages.
```
waab76: Open discussion on the other PR, but I don't think it makes any sense to enforce a max on the limit value.

waab76 (changes requested):

lingpri:
> he @lingpri looks good but will you fix the build error or shall I do?

@jalogisch - I have added a spacing, hopefully, its fixes the build issue.

pages/alerts.rst
```diff
@@ -135,7 +135,10 @@ Event Data
         - ``priority`` (long) - The event's priority value.
         - ``fields`` (Map<String, String>) - The custom fields attached to the event.
 
+.. _Backlog:
+
 Backlog
+-------
```
waab76: Now that you're not linking to this from elsewhere in the documentation, I think you can get rid of the anchor and section heading.

```suggestion
Backlog
```

pages/alerts.rst
```diff
@@ -312,6 +315,9 @@ Configuration Color
 Custom Message
     The message that will be sent to Slack.  The data described above can be used in this template.
 
+Message Backlog Limit (optional)
+    Limit the number of backlog messages sent as part of the Slack notification.
```
waab76: Should add something here to indicate that if the value is set to 0, no limit will be enforced (similar to the UI)

waab76 (changes requested):

pages/alerts.rst
lingpri: I have added the line in sync with UI help block. ` Limit the number of backlog messages sent as part of the Slack notification. If set to 0, no limit will be enforced.`

PR #1004, opened 2020-11-18
Rework Enterprise Setup Page

malcyon:
This is my attempt at reworking the enterprise setup page. Now that we have the `graylog-enterprise` package, we should be able to make the instructions a lot simpler. There is no more need to separately install plugins, since they are included in the package.

I changed a lot, so I totally expect someone to not be okay with things I changed. This is just my stab at it. I stole a lot of stuff from other pages, then trimmed it down to be more concise. Feedback welcome.


malcyon:
@jalogisch How does this look?

pages/enterprise/setup.rst
```diff
@@ -4,46 +4,192 @@
...
 
-Graylog Enterprise comes as a Graylog server plugin which need to be installed
-in addition to the Graylog open source setup.
+Graylog Enterprise comes as a single package that includes Graylog Server and the following plugins:
 
-Requirements
-============
+- Enterprise plugin
+- Integrations plugin
+- Enterprise Integrations plugin
```
jalogisch, line 12: I would add something like:

All parts could be received also as a single package. 

Or similar with links to the integration plugin installation packe and the list with single package names

pages/enterprise/setup.rst
```diff
@@ -4,46 +4,192 @@
...
+- for non-string content in the field, the byte length of that content is counted
+    - byte = 1 byte
+    - char/short = 2 bytes
+    - bool/int/floar = 4 bytes
+    - long/double = 8 bytes
+    - dates = 8 bytes
+- all internal fields are not countent (those meta information that are created by Graylog)
+
+
+Updating to latest version
```
jalogisch: I would move this section above the license verification, that you have install and update directly after each other. 

In addition we might want to break this out into a single page and link this only. Similar what is done in the searching part: https://docs.graylog.org/en/4.0/pages/queries.html

It is for many people harder to read on a long page so having multiple single pages could make updates also easier. 

What did you think?

pages/enterprise/setup.rst
```diff
@@ -57,29 +203,29 @@ When the usage of online repositories is not possible in your environment, you c
...
 DEB
-~~~
+---
 
 The installation on distributions like Debian or Ubuntu can be done with *apt-get* as installation tool from the previous installed online repository.
 
 ::
 
-  $ sudo apt-get install graylog-enterprise-plugins
+  $ sudo apt-get install graylog-enterprise-plugins graylog-integrations-plugins graylog-enterprise-integrations-plugins
```
jalogisch: did you need to update the parts of the meta package or can you update the meta package?

pages/enterprise/setup.rst
```diff
@@ -154,10 +301,10 @@ Your plugin directory should look similar to this after installing the enterpris
...
 
   plugin/
-  ├── graylog-plugin-aws-3.3.2.jar
-  ├── graylog-plugin-collector-3.3.2.jar
-  ├── graylog-plugin-enterprise-3.3.2.jar
-  └── graylog-plugin-threatintel-3.3.2.jar
+  ├── graylog-plugin-aws-3.3.8.jar
+  ├── graylog-plugin-collector-3.3.8.jar
+  ├── graylog-plugin-enterprise-3.3.8.jar
+  └── graylog-plugin-threatintel-3.3.8.jar
```
jalogisch: is this still true?

jalogisch (changes requested):
Please look at my comments.

We can work on that together if you like, but I would like to get your thoughts first.

pages/enterprise/setup.rst
malcyon: Do you mean something like this? 

```
If you're already running an open source instance, you can `install these plugins separately <#update-graylog-instance-to-graylog-enterprise>`_ to convert it to an Enterprise instance. 
```

malcyon: I like the idea of having each plugin name have a hyperlink to more information about it, but I'm not sure what page would contain more info on each plugin.

pages/enterprise/setup.rst
malcyon: Yes, good idea.

pages/enterprise/setup.rst
malcyon: The `graylog-enterprise-plugins` isn't a meta package. If you install it, it'll only install these files:

```
vagrant@instance:~$ dpkg -L graylog-enterprise-plugins
/.
/usr
/usr/share
/usr/share/graylog-server
/usr/share/graylog-server/plugin
/usr/share/graylog-server/plugin/LICENSE-ENTERPRISE
/usr/share/graylog-server/plugin/graylog-plugin-enterprise-4.0.1.jar
/usr/share/graylog-server/bin
/usr/share/graylog-server/bin/chromedriver_start.sh
/usr/share/graylog-server/bin/headless_shell
/usr/share/graylog-server/bin/chromedriver
/usr/share/doc
/usr/share/doc/graylog-enterprise-plugins
/usr/share/doc/graylog-enterprise-plugins/changelog.gz
```

So while installing it does convert the Graylog instance into an Enterprise instance, since we're including the integrations plugins into the enterprise server package, I thought we might as well include them here, too.

Not sure if that's what you meant?

pages/enterprise/setup.rst
malcyon: I think so. I only updated the version here. It's just an example of what a plugins directory might look like.

pages/enterprise/setup.rst
malcyon: I'll break it out into a separate page.

malcyon:
@jalogisch 

I split up the sections into separate pages, but I can't seem to get the sections to show up properly in the sidebar. Currently, they are all nested under the "Starting Graylog" section. What's a good way to fix this?

malcyon:
@jalogisch Any idea why the license page doesn't show up in the sidebar anymore?

jalogisch:
When you clean the build cache @juju2112 you see that license is now on the top level:

image

pages/enterprise/setup.rst
jalogisch: Ok, I thought that you could use `graylog-enterprise` as this is a meta package for installation that contains all of the single packages that we have so that you install one package and have 3 installed. 

That is the reason I though we should refer in the docs only to this meta package as main installation option.

pages/enterprise/setup.rst
jalogisch: In the past we tried to keep that in the same picture what customers can see on their systems. that is why I ask.

jalogisch (dismissed):
as this will be merged in the 4.0 branch we need to add the 4.0 enterprise links and change the versions accordingly but it looks good.

pages/enterprise/setup.rst
malcyon: Oh, I see what you're saying. I think that could work.

malcyon: So then it would delete all the bin files for graylog-server and reinstall them from the `graylog-enterprise` package.

malcyon: Alright, it should be updated now.

malcyon:
> as this will be merged in the 4.0 branch we need to add the 4.0 enterprise links and change the versions accordingly but it looks good.

Can you clarify what you mean? What links do we add where?

jalogisch:
at least in https://github.com/Graylog2/documentation/pull/1004/files#diff-75f182e16ce5544431136c3e0b411ab447dce9a08260be1c489c9cb7af3ab500R52-R95 we are not talking about the 4.0 version of Graylog but this branch is for the 4.0 version of the docs.

malcyon:
> at least in https://github.com/Graylog2/documentation/pull/1004/files#diff-75f182e16ce5544431136c3e0b411ab447dce9a08260be1c489c9cb7af3ab500R52-R95 we are not talking about the 4.0 version of Graylog but this branch is for the 4.0 version of the docs.

Oh! Excellent point. I missed that. I'll fix that, and then I think we're good.

jalogisch (approved):
👍

PR #1055, opened 2021-01-28
Add information on full-message formatter

waab76:
This change adds documentation describing the Full Message Formatter in the Enterprise Output Framework.


dulanism (approved):
+1. This reads fine to me. Thanks @waab76

PR #1057, opened 2021-01-29
Add docs for mounting plugin dir as a volume

malcyon:
Updating the Docker documentation to show mounting the plugin dir as a volume. 

Previously, users had to mount each plugin individually or create their own image using ours as a base. As of 4.0.2, users can mount the `plugin` dir as a volume and it will not override the default plugins anymore. 

https://github.com/Graylog2/graylog2-server/issues/9521
Graylog2/graylog-docker#148

I have also updated the `docker-compose.yml` examples to use the [wait-for-it](https://github.com/vishnubob/wait-for-it) script. This makes docker-compose wait for Elasticsearch to start up before Graylog starts. See https://github.com/Graylog2/graylog-docker/issues/149.

### DIFF
--- a/pages/installation/docker.rst
+++ b/pages/installation/docker.rst
@@ -129,6 +129,7 @@ Example Version 2::
         # Password: admin
         - GRAYLOG_ROOT_PASSWORD_SHA2=8c6976e5b5410415bde908bd4dee15dfb167a9c873fc4bb8a81f6f2ab448a918
         - GRAYLOG_HTTP_EXTERNAL_URI=http://127.0.0.1:9000/
+      entrypoint: /usr/bin/tini -- wait-for-it elasticsearch:9200 --  /docker-entrypoint.sh
       links:
         - mongodb:mongo
         - elasticsearch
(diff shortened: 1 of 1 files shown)


pages/installation/docker.rst
```diff
@@ -399,67 +402,40 @@ Start all services with exposed data directories::
...
 
-New Docker image
-----------------
+    $ mkdir plugin
+    $ wget https://downloads.graylog.org/releases/graylog-integrations/graylog-integrations-plugins-4.0.2.tgz
+    $ tar -xvzf graylog-integrations-plugins-4.0.2.tgz graylog-integrations-plugins-4.0.2/plugin/graylog-plugin-integrations-4.0.2.jar --directory plugin
+    $ docker run --name graylog --link mongo --link elasticsearch \
+        -p 9000:9000 -p 12201:12201 -p 1514:1514 \
+        -e GRAYLOG_HTTP_EXTERNAL_URI="http://127.0.0.1:9000/" \
+        -v <PATH_TO_LOCAL_PLUGIN_DIR>:/usr/share/graylog/plugin \
```
bernd: Since we are creating the plugin directory in this example, I think we can also use it in the docker command:

```suggestion
        -v ./plugin:/usr/share/graylog/plugin \
```

bernd (changes requested):
Thanks for the update. One small nit:

malcyon:
> Thanks for the update. One small nit:

Good point. Thanks!

bernd (approved):

PR #1076, opened 2021-02-17
Update graylog-sidecar packaging instructions.

malcyon:
Removing `yum update` from sidecar install documentation, as that updates every package installed on the system. The install works fine without this command.

### DIFF
--- a/pages/sidecar.rst
+++ b/pages/sidecar.rst
@@ -70,7 +70,7 @@ CentOS
 Install the Graylog Sidecar repository configuration and Graylog Sidecar itself with the following commands::
 
     $ sudo rpm -Uvh https://packages.graylog2.org/repo/packages/graylog-sidecar-repository-1-2.noarch.rpm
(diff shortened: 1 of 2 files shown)


pages/sidecar.rst
```diff
@@ -70,7 +70,7 @@ CentOS
 Install the Graylog Sidecar repository configuration and Graylog Sidecar itself with the following commands::
 
     $ sudo rpm -Uvh https://packages.graylog2.org/repo/packages/graylog-sidecar-repository-1-2.noarch.rpm
-    $ sudo yum update && sudo yum install graylog-sidecar
```
bernd, line 73: Can you remove the `yum update` from the `pages/installation/os/centos.rst` as well? Thanks!

bernd (changes requested):

pages/sidecar.rst
malcyon: Sure thing!

bernd (approved):
Thank you! :+1:

PR #1083, opened 2021-02-22
Hiding company name in header by using CSS (4.0).

linuspahl:
## Description

With https://github.com/Graylog2/documentation/pull/1064/files#diff-6e03d3d89630aba3403afea3514a3ac5948e70d1ced091012e56446f614da340R123 we tried to display the company logo in the sidebar header only. Unfortunately the config `html_theme_options` have no effect on the production system.

With this PR we are hiding the company name by using CSS.


jalogisch (approved):
should work now

PR #1097, opened 2021-02-26
Hiding company name in header by using CSS (3.3).

linuspahl:
## Description

With https://github.com/Graylog2/documentation/pull/1065 we tried to display the company logo in the sidebar header only. Unfortunately the config `html_theme_options` have no effect on the production system.

With this PR we are hiding the company name by using CSS.


dulanism (approved):
I like this improvement, too! +1

PR #1100, opened 2021-03-03
Update Sidecar requirements documentation.

malcyon:
I have not released Graylog Sidecar `1.1.0` yet, but when I do, I'd like the documentation to be clear on the new server requirements (See https://github.com/Graylog2/collector-sidecar/pull/391). 

This should not be merged yet (not until Sidecar `1.1.0` is released). But I would like feedback on how clear this looks. Does it make sense? 

I think the idea is that if Sidecar installs Beats 7.0, that older versions of Graylog won't work with it.


pages/sidecar.rst
```diff
@@ -28,15 +28,17 @@ You can get .deb and .rpm packages for Graylog Sidecar in our package repository
...
-| 1.0.x           | 3.0.x,3.1.x,3.2.x                |
-+-----------------+----------------------------------+
-| 0.1.x           | 2.2.x,2.3.x,2.4.x,2.5.x,3.0.x    |
-+-----------------+----------------------------------+
-| 0.0.9           | 2.1.x                            |
-+-----------------+----------------------------------+
++-------------------+-----------------------------------+
+| Sidecar version   | Graylog server version            |
++===================+===================================+
+| 1.1.x             | 3.2.5 or higher, 4.0 or higher    |
```
mpfz0r: Graylog 3.3.x also supports the new beats collectors.
I'm not sure how explicit we should be here.
I'd say `3.2.5 or higher` covers everything

mpfz0r:
> I think the idea is that if Sidecar installs Beats 7.0, that older versions of Graylog won't work with it.

Just to clarify: 
This isn't really a sidecar incompatibility, but a beats one.
It only affects the Windows sidecar, because that's the only one where we are shipping the beats binaries.
On Linux, users are responsible themselves to install beats. And they could've already updated beats to 7.x  and experienced problems.

And I'm not suggesting we should document any of that ;-)

pages/sidecar.rst
malcyon: Good point!

malcyon:
@mpfz0r Thanks for the clarification!

mpfz0r (approved):
Looks good. I guess we should also apply that change to the older doc branches. like at least 3.3 and 3.2.

malcyon:
> Looks good. I guess we should also apply that change to the older doc branches. like at least 3.3 and 3.2.

Right. I'll take care of that, too.

dulanism (approved):
+1. Thank you, @juju2112

PR #1117, opened 2021-03-22
Warn users not to run graylog with elastic 7.11

dulanism:


pages/installation.rst
```diff
@@ -33,6 +33,8 @@ This chapter is explaining the many ways to install Graylog and aims to help cho
 System requirements
 ===================
 
+.. warning:: We recommend that you avoid Elasticsearch 7.11; it will cause your instance to break.
```
lennartkoopmann: Can we make this a little clearer and say that you absolutely shouldn't because it *will* break?

Is this the only place in the docs where we mention Elasticsearch versions?

dulanism:
@lennartkoopmann Take a look at these changes. I wanted to throw in some more urgency. Lemme know if it needs to be softened.

lennartkoopmann:
Thank you! I think the Operating System Packages still mention it and, to make it worse, even say "7.x": https://docs.graylog.org/en/4.0/pages/installation/operating_system_packages.html#deb-apt

dulanism:
Fixed cc @lennartkoopmann

pages/configuration/elasticsearch.rst
```diff
@@ -14,6 +14,7 @@ Elasticsearch versions
 
 Starting with version 2.3, Graylog uses the HTTP protocol to connect to your Elasticsearch cluster, so it does not have a hard requirement for the Elasticsearch version anymore. We can safely assume that any version starting from 2.x is working.
 
+.. warning:: We caution you not to upgrade Elasticsearch to 7.11! You are absolutely forbidden to upgrade to this version. If you do so, it will break your instance.
```
bernd: Use "!" at end of sentence for consistency:
```suggestion
.. warning:: We caution you not to upgrade Elasticsearch to 7.11! You are absolutely forbidden to upgrade to this version. If you do so, it will break your instance!
```

pages/installation/operating_system_packages.rst
```diff
@@ -22,6 +22,8 @@ The repositories can be set up by installing a single package. Once that's done,
 Prerequisites
 -------------
 
+.. warning:: We caution you not to upgrade Elasticsearch to 7.11! You are absolutely forbidden to upgrade to this version. If you do so, it will break your instance.
```
bernd: Use "!" at end of sentence for consistency:
```suggestion
.. warning:: We caution you not to upgrade Elasticsearch to 7.11! You are absolutely forbidden to upgrade to this version. If you do so, it will break your instance!
```

pages/upgrade.rst
```diff
@@ -37,6 +37,8 @@ For .rpm based systems :ref:`this update guide <operating_package_upgrade_rpm-yu
 Upgrading Elasticsearch
 -----------------------
 
+.. warning:: We caution you not to upgrade Elasticsearch to 7.11! You are absolutely forbidden to upgrade to this version. If you do so, it will break your instance.ß
```
bernd: Use "!" at end of sentence for consistency and remove stray "ß" character:
```suggestion
.. warning:: We caution you not to upgrade Elasticsearch to 7.11! You are absolutely forbidden to upgrade to this version. If you do so, it will break your instance!
```

bernd (changes requested):
Some suggestions:

pages/upgrade.rst
dulanism: Take a look now, @bernd. Good to go?

bernd: @dulanism Thanks for the fixes! :+1: 

To be honest, the wording sounds rather strange to me. I am not a native speaker, so please take my comment with a grain of salt! :smile: 

I am specifically talking about the "You are absolutely forbidden to upgrade to this version." part. That sounds very strong and weird, because we cannot forbid a user to do something. :smile: My personal preference would be something along the lines of:

> We caution you not to install or upgrade Elasticsearch to 7.11 and later! It is not supported. If you do so, it will break your instance!

What do you think?

bernd (approved):
Thank you! LGTM :+1:

lennartkoopmann:
I found a few more pages on which we say that Graylog is compatible with "Elasticsearch 7.x", which is now no longer true:

* https://docs.graylog.org/en/4.0/pages/installation/operating_system_packages.html
* https://docs.graylog.org/en/4.0/pages/installation/os/ubuntu.html
* https://docs.graylog.org/en/4.0/pages/installation/os/sles.html
* https://docs.graylog.org/en/4.0/pages/installation/os/debian.html
* https://docs.graylog.org/en/4.0/pages/installation/os/centos.html

I suggest that we clarify on those pages. @dulanism, should we consider having one page that lists all compatibilities and link to it instead of having to maintain all those pages individually?

dulanism:
@lennartkoopmann Aye Karumba! Create a whole page dedicate to compatibilities? Yeah. That's a great idea. Before I do a PR here, lemme create a sheet to get an idea of how it will look in the docs.

PR #1159, opened 2021-06-10
Update pages for 4.1

bernd:
This updates the documentation for 4.1.

Previously we had to go through all pages and replace the last version with the current version in code blocks because Sphinx doesn't support variables in code blocks. We now register a custom `source-read` function that does the replacement. All pages have been updated to use the placeholders. (e.g. `{{SERVER_VERSION}}`)

Next time we only have to update the version number in `conf.py`.


dulanism (approved):
Thanks for doing this @bernd. Good to go @juju2112 +1

PR #1160, opened 2021-06-15
Type conversion for message field values

jrunu:
Explicit mention that message field values from a message should always be casted before used. This can
be inferred by all the rule examples in the documentation, but is never explicitely stated.
 
While there is a case with strings, where it works without an explicit cast, this is not considered valid
according to an exchange with the Graylog Support (#446984520). Hence we're proposing to add
the following note as clarification/best practice.

## Notes for Reviewers

- [ ] The commit history must be preserved - please use the rebase-merge or standard merge option instead of squash-merge
- [ ] Sync up with the author before merging


bernd (approved):
Thank you! :pray:

PR #1166, opened 2021-06-17
Add Prometheus Exporter docs

### LINKED ISSUE #1154: Document new core Prometheus Exporter
In the 4.1 release, the ability to export Graylog Dropwizard metrics to Prometheus is being added to the core open-source product in https://github.com/Graylog2/graylog2-server/pull/10767. This works similarly to the existing Graylog Cloud server-side prometheus metrics exporting, except it allows the mappings between the Dropwizard and Prometheus metrics to be customized at runtime in a yaml file. 

## Expected Content

* Add the new configuration parameters to the [server.conf](https://docs.graylog.org/en/4.0/pages/configuration/server.conf.html?highlight=server.conf) docs page.
* Add a new page or section that describes the minimum details needed to understand, enable, and use the feature. There is no need to explain how to set up Prometheus. However, we should explain how to define the individual mappings.

## Functionality Description

The configuration properties described in the `server.conf` file can be used to enable and customize the feature. 

This Prometheus exporter heavily uses [Prometheus metric labels](https://prometheus.io/docs/practices/naming/) to consolidate individual traditional Graylog Dropwizard metrics in to a combined metric with labels, which make it easer to view and analyze similar metrics together.

The initial export mappings are described here in [prometheus-exporter.yml](https://github.com/Graylog2/graylog2-server/blob/master/graylog2-server/src/main/resources/prometheus-exporter.yml). Essentially, this is a mapping between Dropwizard metric names (wildcard patterns) and prometheus export names. The matching patterns are described in detail [here](https://github.com/prometheus/client_java#dropwizardexports-collector). 

The following example definition uses most of the export, mapping and labels functionality, note the `metric_name` and `metric_pattern`. All metrics exported to Prometheus will have the `gl_` prefix appended to the name. So, the name in the below example will result as `gl_auth_actions`. 

```
  - metric_name: "auth_actions"
    match_pattern: "org.graylog.plugins.cloud.auth.*.failure.http.*"
    wildcard_extract_labels:
      - "action"
      - "http_code"
    additional_labels:
      success: "false"
      type: "http"
```

The `wildcard_extract_labels` are an ordered sequence of label names to extract for the wildcard asterisks `*` that appear in the metric name. One label value for each value in place of the wildcard will be created.
 
For example, for the definition above, the following metric would produce the following labels: 

`metric: org.graylog.plugins.cloud.auth.tokenRequest.failure.http.500`

`label: action: tokenRequest`
`label: action: http_code: 500`

The `success` and `type` labels are static labels.

Note that all metrics will automatically be assigned a `node_id` label with the value of the Graylog node id that the metric was exported from. This makes it easy to visualize metrics across Graylog nodes together, and also see which values are coming from which node.

I hope this explanation helps, but perhaps the best way to see how to definitions work is to use the exporter for some known metrics and map them to prometheus metrics and various labels to see what the output looks like.

I am not exactly sure how much detail we should go into in the documentation, since to use this feature, the user probably already knows Prometheus fairly well. 

To test this functionality, the instructions to starting a Prometheus instance in Docker can be used:
https://docs.graylog.org/en/4.0/pages/enterprise/cloud_forwarder.html#prometheus-metrics-exports

danotorrey:
Add documentation for the new core Prometheus Exporter feature (originally implemented in https://github.com/Graylog2/graylog2-server/pull/10767).

We did not yet have a main area in the docs where the Metrics topic is covered, so I added one so I have added a new top-level page titled `Metrics`. I added a general description of standard Graylog metrics at the top as a prelude to discussing Prometheus exports in detail. 

A bulk of the Metrics page described the new Prometheus exporting functionality. It ended up a bit more extensive than I was originally thinking, but I think most of the content will be useful for those who really want to fully utilize the feature. 

This was written in a rush, so hopefully this all makes sense :D 

Closes #1154


danotorrey:
@bernd @dulanism Since this will be included in the 4.1 release, and some users are wanting to utilize it, I wanted to add some docs for it in time.

If you can find the time to do an initial review of this tomorrow (Friday) or Monday morning, that would be really helpful. I would like to try and merge this on Monday so that it will be ready to go on release day. Thanks!

pages/metrics.rst
```diff
@@ -0,0 +1,159 @@
...
+    http://127.0.0.1:9833/api/metrics/prometheus
+
+To specify a custom export hostname or IP address, the following configuration can be used.
+We suggest leaving the default export port `9833`, since it is already `registered with Prometheus <https://github.com/prometheus/prometheus/wiki/Default-port-allocations>`_.
+
+.. code-block::
+
+    prometheus_exporter_bind_address = 10.0.0.1:9090
+
+Depending on the current Prometheus `scrape_config setting <https://eus.io/docs/prometheus/latest/configuration/configuration/#scrape_config>`_,
```
bernd: The link has been truncated:

```suggestion
Depending on the current Prometheus `scrape_config setting <https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config>`_,
```

pages/metrics.rst
```diff
@@ -0,0 +1,159 @@
...
+the Graylog Prometheus exporter may be discovered automatically, or an explicit scrape target might need to be
```
bernd: How would the exporter be discovered automatically? Is there a way to do that? :slightly_smiling_face:

bernd (changes requested):
Thank you! :pray: Only one comment and one question:

pages/metrics.rst
danotorrey: Thanks for seeing this. From a second-look, I think I misunderstood how the automatic discovery works. It looks like this is only supported in some special configurations (such as [Kubernetes setups](https://developer.sh/posts/prometheus-self-discovery)) where the list of cluster host names is already known through other means. It's not really applicable for the purposes of this docs page. 

I have re-worded this section to be a bit more general and not reference auto-discovery.

bernd (approved):

PR #1174, opened 2021-06-22
4.1 Forwarder documentation

### LINKED ISSUE #1145: 4.1 Forwarder Documentation
Modify the [existing Cloud Forwarder documentation](https://docs.graylog.org/en/4.0/pages/enterprise/cloud_forwarder.html) to target the new 4.1 Forwarder.

In the upcoming 4.1 release of Graylog, we will be migrating the Cloud Forwarder and making it available in the core on-premise Graylog Server product (Enterprise customers only). Currently, we are just calling this the “Forwarder”. 

So, starting in Graylog 4.1, we will maintain this one Forwarder tool that can be used for both Cloud and on-premise Graylog installations. The main benefit for on-premise customers, is that they can forward data from other networks or locations to their central Graylog installation. The functionality remains the same for Cloud installations.

So, with the new 4.1 Forwarder, we will need to modify the [existing Forwarder documentation](https://docs.graylog.org/en/4.0/pages/enterprise/cloud_forwarder.html) to describe the general 4.1 Forwarder instead.

A few specifics on the modifications needed to the [existing Forwarder docs](https://docs.graylog.org/en/4.0/pages/enterprise/cloud_forwarder.html):
- Product name: The name will change from "Cloud Forwarder" to just "Forwarder"
- Most of the implementation of the Forwarder (how it works internally) will remain the same. 
- We will release one version of the forwarder that works with both Cloud and on-premise. So the installation (tar, os packages, docker should work for both). I think the OS packages/repos should remain mostly the same. 
- The Forwarder will be versioned with a simple Major.Minor version scheme (eg. 4.1, 4.2). So, starting out, we will release 4.0. The major version will always be compatible with the same major version of Graylog. More info about Forwarder [versioning here](https://graylogdocumentation.atlassian.net/wiki/spaces/GENERAL/pages/1650982944/Standalone+Forwarder#Versioning). We should include an explanation about the compatibility of versions in the docs.
- Since the Forwarder will be compatible with Graylog Cloud and Graylog On-premise, I think the documentation will need to provide instructions for both setups. The only thing that is different will be how they are configured. For on premise, they will need to provide their own server hostname/IP and TLS certs.
- TLS is automatic with the existing Cloud Forwarder, and will continue to be for Cloud in the new Forwarder. But, for on-premise, the customer will need to provide their own cert and key on the server/client side. See the cert and key properties in the [wiki](https://github.com/Graylog2/graylog-plugin-cloud/wiki/Forwarder#forwarder-configuration-properties-client-side). `forwarder_grpc_tls_trust_chain_cert_file` in the Forwarder and `forwarder_grpc_tls_private_key_file` and `forwarder_grpc_tls_trust_chain_cert_file` on the server-side.
- The Forwarder setup wizard will change slightly with the new 4.1 Forwarder. It will need to be able to provide the configuration for both Cloud and on-premise settings depending on which is running. We need to look at this more to understand what it will look like.

Development on the 4.1 Forwarder is still in-progress, and as it continues, we will probably find a few more details to include in the documentation.

danotorrey:
Migrates existing Cloud Forwarder documentation to cover the new 4.1 Forwarder, which can be used for on-premise and Cloud.

The goal of this new documentation was to provide a single reference for the new 4.1 Forwarder, and also provide the specific configurations for both Cloud and on-premise.

Closes #1145


dulanism:
Great collaboration @danotorrey! 

There were only two items that I recommend updating

1. Can you update the Forwarder selection image to the item attached? 
forwarder_select

2. The table at the bottom is completely deprecated. You can toss it.

Aside for those items is production-ready IMHO :)

dulanism (dismissed):
+1 after minor suggestions are handled. Thanks again!

PR #1173, opened 2021-06-22
Initial documentation for URLhaus data adapter

waab76:
This PR adds documentation for the new URLhaus enterprise data adapter.

Doc entry in the sidebar:
Screen Shot 2021-06-22 at 2 31 04 PM

Doc entry on the Integrations page:
Screen Shot 2021-06-22 at 2 31 26 PM

Documentation:
Screen Shot 2021-06-22 at 2 22 05 PM
Screen Shot 2021-06-22 at 2 22 19 PM

## Notes for Reviewers

- [ ] The commit history must be preserved - please use the rebase-merge or standard merge option instead of squash-merge
- [ ] Sync up with the author before merging

PR #1179, opened 2021-06-23
Fix URLHaus Feed Type indenting

danotorrey:
Fix the indenting, which was done improperly by @danotorrey in attempt to fix the PR check failure. https://github.com/Graylog2/documentation/pull/1173#issuecomment-866854748

It looks like Sphinx did not like the varied indent that was originally coded in #1173 

> Warning: Definition list ends without a blank line; unexpected unindent.

I attempted to fix it in https://github.com/Graylog2/documentation/pull/1178/commits/1cd2b0756e3de1a78de02e6b81165d500e05553f, but I actually made the indent correct. 

This was the best I could do to fix it.

image

### DIFF
--- a/pages/integrations/lookups/urlhaus.rst
+++ b/pages/integrations/lookups/urlhaus.rst
@@ -45,9 +45,9 @@ Data Adapter Configuration
 - ``Custom Error TTL``
    - Optional custom TTL for caching erroneous results.  If no value is specified, the default value of 5 seconds will be used.
 - ``URLhaus Feed Type``
-	- This determines which URLhaus feed the data adapter will use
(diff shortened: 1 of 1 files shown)


danotorrey:
@waab76 Please let me know if there is a better way to format the indent. I think the original idea was for the Feed Type choices to be indented further, but I don't see a way for that to work in a way that is approved by the PR checks.

dulanism (approved):
Just pulled down the branch, inspected it, and it looks good! +1 cc @waab76

waab76 (approved):
This looks good.  Thanks for the fix!

PR #1198, opened 2021-07-23
Add notes about TLS cipher changes in 4.1

bernd:
This adds a few clarifications about TLS cipher changes in 4.1 because some people ran into issues.


mpfz0r (approved):
LGTM :+1: and thank you for that  :)

PR #1217, opened 2021-09-22
Add bugfix release schedule info to FAQ.

malcyon:
Adding info on when we do bugfix releases to FAQ.


dulanism (approved):
Looks good to me. Thanks @juju2112

